PR #12, opened 2015-11-08
style(index) disabling irc link for mobile media query

jackfrancis:
iOS/Safari at least, and I'd expect most mobile browser implementations, doesn't do anything with a `irc://`-prefixed hyperlink.


jackfrancis:
@slack proposing to just remove the IRC button from the mobile site until we get a browser-friendly implementation of freenode#helm

PR #38, opened 2017-05-09
Copy and link updates to Helm.sh

flynnduism:
* link to new Docs (helm.sh/docs)
* update maintainer info (add CNCF, Deis to Microsoft)
* link the Get Charts to monocular (kubeapps.com)


flynnduism:
Thanks, links updated.

PR #46, opened 2017-05-26
Apply the latest docs edits from source

flynnduism:
This commit rounds up and ports over any edits to the Helm Docs at [the source](https://github.com/kubernetes/helm/tree/master/docs), since the initial setup of this site.

--  

The content should be automatically pulled from `kubernetes/helm`.  [#2501](https://github.com/kubernetes/helm/pull/2501) is the first step, so that the source will follow the format of this site - then we can update the build script to pull directly from there.


technosophos:
LGTM

Issue #51, opened 2017-07-07
"Using Helm" docs section not built properly?


dlapiduz:
Browsing to https://docs.helm.sh/using_helm/#quickstart-guide should render the quickstart guide but right now I only see the "Using Helm" subsection of the "Using Helm" section.

I wonder if this is a build issue because running hugo on the docs folder renders the section just fine.
![screen shot 2017-07-07 at 8 53 15 am](https://user-images.githubusercontent.com/4604/27960799-328541be-62f2-11e7-9d68-ca6d2084810b.png)
![screen shot 2017-07-07 at 8 54 23 am](https://user-images.githubusercontent.com/4604/27960798-327deca2-62f2-11e7-91ff-9e519ac2d703.png)

flynnduism:
Yep, looks an issue with the build sequence - part of the build never ran. Thanks for flagging this issue. 

I've a fix on the way asap.

flynnduism:
Fixed and deployed.

That content is accessible again at: [https://docs.helm.sh/using_helm/#quickstart-guide](https://docs.helm.sh/using_helm/#quickstart-guide)

PR #52, opened 2017-07-07
Update build scripts

flynnduism:
There's an issue with `gulp` tasks sequencing that's [affecting some builds (#51)](https://github.com/helm/helm-www/issues/51), where the directory setup does not complete.

This PR splits the gulp tasks so that the directory setup waits for the fetching of the docs to complete before running, to ensure the correct sequence is followed.

### DIFF
--- a/docs/gulpfile.js
+++ b/docs/gulpfile.js
@@ -256,11 +256,22 @@ gulp.task('clone', function() {
   });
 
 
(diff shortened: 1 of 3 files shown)


vdice:
LGTM -- haven't checked staging env, but if it looks good there, then 👍 !

PR #55, opened 2017-07-12
Update docs menu to fix #49 and #54

### LINKED ISSUE #49: [docs] make the home link relative to site baseurl
Related to #47, suggestion from @vdice.

The side menu home link is currently set to `docs.helm.sh`, using a hard-coded link instead of a relative parameter from Hugo.

Currently, the `site.baseurl` param can't be used in `config.toml`, but there's likely another easy way to make this a relative link so it works better in local development and staging.

### LINKED ISSUE #54: Helm documentation hyperlinks are dangling
From the Helm project's [root _README.md_ file](https://github.com/kubernetes/helm/blob/master/README.md), follow [the link to the Quick Start guide](https://docs.helm.sh/using_helm/#quickstart-guide). We arrive at the intended anchor on the intended page, but from there **several of the hyperlinks in the left sidebar (revealed via the blue-striped "hamburger menu") don't work**.

It looks like they mistakenly include the URL path component "using_helm" when they should not.
- The "Using Helm" link is `https://docs.helm.sh/using_helm/using_helm`, which yields a 404 response.
- The "Helm Commands" link is `https://docs.helm.sh/using_helm/helm`, which yields a 404 response.
- The "Charts" link is `https://docs.helm.sh/using_helm/developing_charts`, which yields a 404 response.
- The "Developing Templates" link is `https://docs.helm.sh/using_helm/chart_template_guide`, which yields a 404 response.
- The "Best Practices" link is `https://docs.helm.sh/chart_best_practices/` which doesn't get a response in a reasonable amount of time. (It came through after a couple of minutes.)
- The "Relate Projects" link is `https://docs.helm.sh/related/` which doesn't get a response in a reasonable amount of time. (It came through after a couple of minutes.)

I'm not sure if this is related to the recent kubernetes/helm#2645.

flynnduism:
* update the offcanvas template to fix the docs menu for mobile and small screens (closes #54) 
* remove the home link from the menu (config.toml)
* add the home link into theme templates, as a relative `site.BaseUrl` value (closes #49)

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -19,11 +19,6 @@ defaultContentLanguage = "en"
 
 
 [[menu.main]]
-  name = "Home"
-  url = "https://docs.helm.sh"
-  identifier = "home"
-  weight = 1
-[[menu.main]]
   name = "Using Helm"
   url = "using_helm"
   identifier = "using_helm"
--- a/docs/themes/helmdocs/layouts/partials/offcanvas.html
+++ b/docs/themes/helmdocs/layouts/partials/offcanvas.html
@@ -1,16 +1,25 @@
 <aside class="left-off-canvas-menu" id="off-canvas-navigation" data-topbar>
   <div class="drawer-wrap">
 
-  <ul class="off-canvas-list">
-      <li><a href="{{ .Site.BaseURL }}">Helm Docs Home</a></li>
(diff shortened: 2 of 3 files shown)


flynnduism:
Deployed to [staging](http://docs-helm-sh.staging.deis.com/)

PR #77, opened 2018-03-08
Helm Site Housekeeping

flynnduism:
* remove old files from version control
* remove Hugo compiled files from `docs.helm.sh/app`
* bump npm packages to secure version
* remove npm packages that are breaking the `make build` deploy script
* update `.gitignore`


bacongobbler:
Looks like there might be a few extras that can be removed in https://github.com/helm/helm-www/tree/master/docs.helm.sh/rootfs/public. That looks like it's the site generated after hugo's done its thing. Would you mind removing that as well?

flynnduism:
Good catch @bacongobbler thanks!

PR #87, opened 2018-04-26
use root baseurl

### LINKED ISSUE #86: CORS issue on helm.sh
Reported in Slack:

```
Access to Font at 'https://www.helm.sh/assets/fonts/klinicslabmedium-webfont.woff2' from origin 'https://helm.sh' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://helm.sh' is therefore not allowed access.
```

bacongobbler:
closes #86.

### DIFF
--- a/docs.helm.sh/config.toml
+++ b/docs.helm.sh/config.toml
@@ -1,4 +1,4 @@
-baseurl = "https://docs.helm.sh/"
+baseurl = "/"
 languageCode = "en-us"
 title = "Helm Docs"
 theme = "helmdocs"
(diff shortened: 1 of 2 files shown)


bacongobbler:
hmm, it looks like helm.sh is being built using Jekyll rather than Hugo, so this might end up different than how I expect it to work.

bridgetkromhout:
Does this still need to happen, given the site updates?

bacongobbler:
now that the codebase has consolidated into a single site, this is no longer necessary. Closing!

PR #89, opened 2018-05-14
update outdated packages

flynnduism:
This bumps the version of older packages on both `helm.sh` and `docs.helm.sh`. 

One of the sub-dependencies used has a known vulnerability, so this resolves that issue and updates any other outdated npm packages (I used [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) for simplicity).


docs.helm.sh/package.json
```diff
@@ -10,37 +10,37 @@
     "url": "git+https://github.com/helm/helm-www"
   },
   "devDependencies": {
-    "del": "^2.2.0",
+    "del": "^3.0.0",
     "font-awesome": "^4.7.0",
-    "foundation-sites": "^5.5.3",
+    "foundation-sites": "5.5.3",
```
flynnduism: I had to make an exception for `foundation-sites`, as bumping this to the latest package version of the framework breaks the appearance, as v6 uses an entirely different set of HTML and CSS.

bacongobbler: I thought the caret operator was supposed to limit the package to the latest minor release, so `>= 5.5.3, < 6.0.0`. Is this not the case?

What about limiting this to the latest patch release via `~5.5.3`?

docs.helm.sh/package.json
```diff
@@ -10,37 +10,37 @@
...
+    "gulp-sourcemaps": "^2.6.4",
+    "gulp-string-replace": "^1.0.0",
     "gulp-tap": "^1.0.1",
     "gulp-track-filenames": "^0.1.0",
-    "gulp-uglify": "^1.5.1",
+    "gulp-uglify": "^3.0.0",
     "gulp-util": "^3.0.8",
     "jshint": "^2.9.4",
-    "remove-markdown": "0.1.0",
+    "remove-markdown": "0.2.2",
```
bacongobbler: Same here, can we do a patch release comparison (`~0.2.2`)?

flynnduism:
Thanks @bacongobbler, updated with `^` and `~` - makes more sense, as you point out!

Issue #92, opened 2018-05-31
docs(bug): rbac url in quickstart 404s


flynnduism:
The docs site renders the RBAC section as `https://docs.helm.sh/using_helm/#role-based-access-control`, but there is url in the Quick Start Guide that points to `/rbac`. 

![rbac-404](https://user-images.githubusercontent.com/686194/40804919-8534d8ac-64d1-11e8-8745-51b8822acb82.png)

This works fine when browsing the docs on github.com/kubernetes/helm, but doesn't work on the Hugo rendered version of the docs site here. Thanks to @slack for flagging this.

This is easily remedied via the build pipeline, which runs some find and replace tasks to adapt the markdown files to work with Hugo.

bacongobbler:
closed via #168

PR #93, opened 2018-05-31
Update the gulp build task

### LINKED ISSUE #91: bug(docs): site content layout is broken
@slack flagged an issue with the docs site not rendering properly, looks like the build task didn't fully run and the page is a mess. Half of the `Using Helm` subsections render correctly, but then the content column ends prematurely.

docs-broken

The second half of the page content is injected at the end of the page as raw html just before the closing </body> tag.

screen shot 2018-05-31 at 12 01 00 pm

---  

To fix this I'm reviewing the build task sequence and enforcing a delay to ensure everything completes properly. It's likely the `git clone` task that fetches the doc content from the Kubernetes/Helm repo is messing up the task order.

flynnduism:
This fixes #91 and #92.


docs.helm.sh/gulpfile.js
```diff
@@ -243,8 +244,12 @@ gulp.task('clone', function() {
 
   gulp.task('redirect-anchor', function() {
     return gulp.src('source/docs/**/*.md')
-      // update installation guides
+      // update quickstart and install links
       .pipe(replace(/\]\(.*install\.md/, '](../using_helm/#installing-helm'))
+      .pipe(replace('#Install-Helm', '#installing-helm'))
+      .pipe(replace('#quickstart]', '#quickstart-guide]'))
```
bacongobbler: out of curiosity, are these closing brackets intentional?

flynnduism: Yeah, I want to replace a specific instance of #quickstart in a markdown url, without impacting on any correct links which would otherwise get rewritten and broken as follows:

```
without limiting with a closing ']'

#quickstart         ->   #quickstart-guide
#quickstart-guide   ->   #quickstart-guide-guide
```

Issue #103, opened 2018-06-13
[docs] brigade.js not writing to blob storage


flynnduism:
The changes I made in #102 aren't resolving the deployment issues to blob storage. Brigly completes the site build, but the files on the storage + CDN aren't changing. 

---

You can see the logs of the [brigly builds here](https://example-api.technosophos.me/v1/project/brigade-9afa37b976ab4bd116557b613000959d24feb08dfac37fcf6ae3cb/builds) - here's an an [example](https://example-api.technosophos.me/v1/job/helmdocs-01cfx7v407pdgmmeeefha7kygp/logs) of one that ran and returned some errors.

@technosophos you may have a diagnosis based on the below

```
Running\n[brigade:k8s] default/helm-sh-01cfx7vg2abb3gnencm2fam69m phase Failed\n  
\u001b[0m\u001b[97m\u001b[41mError\u001b[0m\u001b[90m:\u001b[0m\u001b[37m \u001b[0m\u001b[97mPod helm-sh-01cfx7vg2abb3gnencm2fam69m failed to run to completion\u001b[0m\n
\u001b[0m  \u001b[0m\n
\u001b[0m  \u001b[0m\u001b[90m-\u001b[0m \u001b[0m\u001b[93mk8s.js\u001b[0m
\u001b[90m:\u001b[0m\u001b[93m417\u001b[0m\u001b[37m \u001b[0m\u001b[37mk.readNamespacedPod.then.response\u001b[0m\n
\u001b[0m  \u001b[0m  \u001b[0m\u001b[90m./dist/k8s.js:417:32\u001b[0m\n
\u001b[0m  \u001b[0m\n\u001b[0m\n
[brigade:app]   \u001b[0m\u001b[97m
Error\u001b[0m\u001b[90m:\u001b[0m\u001b[37m \u001b[0m\u001b[97m
job helm-sh(helm-sh-01cfx7vg2abb3gnencm2fam69m): Error: Pod helm-sh-01c
\u001b[0m  \u001b[0m\u001b
[97mfx7vg2abb3gnencm2fam69m failed to run to completion\u001b[0m\n
\u001b[0m  \u001b[0m\n\u001b[0m  \u001b[0m\u001b[90m-\u001b[0m \u001b[0m\u001b[93mbrigadier.js\u001b[0m\u001b[90m:\u001b[0m\u001b[93m56\u001b[0m\u001b[37m \u001b[0m\u001b[37mjr.run.catch.err\u001b[0m\n
[...]

PR #104, opened 2018-06-13
Update dependencies

flynnduism:
For #103, this bumps the ruby version and updates some npm packages - aiming to resolve errors in the Brigade build and deploy of the site.


flynnduism:
This would replace #89 too which has fallen behind.

PR #106, opened 2018-06-15
[helm.sh] update ruby image and version

flynnduism:
This is a follow up to #105 - switching the docker image for helm.sh from node to ruby.

### DIFF
--- a/brigade.js
+++ b/brigade.js
@@ -31,7 +31,8 @@ events.on("exec", (e, p) => {
   buildHelmSh.timeout = timeout
   buildHelmSh.storage.enabled = true
   buildHelmSh.tasks = [
-    "apt-get update -y && apt-get install -yq ruby ruby-dev",
+    "curl -sL https://deb.nodesource.com/setup_9.x | bash -",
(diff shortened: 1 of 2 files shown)


brigade.js
```diff
@@ -27,11 +27,11 @@ events.on("exec", (e, p) => {
   ];
 
   // This job renders helm.sh
-  const buildHelmSh = new Job("helm-sh", "node:9");
+  const buildHelmSh = new Job("helm-sh", "ruby:2.3");
   buildHelmSh.timeout = timeout
   buildHelmSh.storage.enabled = true
   buildHelmSh.tasks = [
-    "apt-get update -y && apt-get install -yq ruby2.2 ruby2.2dev",
+    "apt-get update -y && apt-get install -yq ruby ruby-dev",
```
adamreese: Now that you are starting with ruby, you need to change this line to install node

I think this should work:
```
curl -sL https://deb.nodesource.com/setup_9.x | bash -
apt-get install -y nodejs
```

flynnduism: Thanks @adamreese, done!

PR #108, opened 2018-07-05
Updating the hub to the new location

mattfarina:


mattfarina:
@bacongobbler @flynnduism what does it take for this to be deployed? Is there a visible pipeline somewhere?

flynnduism:
Currently we've got a slackbot that deploys on demand - but that should be more accessible and auto-deploy on merge to master like this. 

Deployment is via [Brigade](https://github.com/azure/brigade), we can update the script to automate.

dankohn:
CNCF has an account with Netlify, which does automated deployments and also automated previews of PRs. We're happy to set you up with a sub-account if you want to switch.

flynnduism:
@dankohn thats good to know - I really like Netlify for static site projects. 

We're not looking to change our kubernetes-centric hosting/tooling, but the [readme needs updating](https://github.com/helm/helm-www/pull/110/) and the resources should be automated and accessible to all core personnel, we've some adjustments to make

PR #123, opened 2018-07-24
feat(go-imports): configure vanity go import path

adamreese:
Enable `go get helm.sh/helm/...` using http.gopkg caddy plugin.
Docs at https://caddyserver.com/docs/http.gopkg


adamreese:
NOTE: This is untested....

technosophos:
That would be cool!

PR #127, opened 2018-07-24
docs scroll fix

flynnduism:
A bug arose following #121 and #122; a css conflict between the footer addition and the overflow edit.

This edit unlocks the viewport so scrolling is corrected.


bacongobbler:
I notice the CNCF logo is being removed in this PR. Was that unintended?

PR #134, opened 2018-08-16
Fixing typos

bridgetkromhout:
I noticed a "to" that is probably supposed to be a "for", and fixed a little punctuation while I was sending a PR anyhow.


mattfarina:
Thanks for cleaning up our comma usage. Since I'm an Oxford comma user I appreciate this. Better merge it before @flynnduism or @technosophos decide to take a stance the other way.

Issue #146, opened 2018-10-16
[helm.sh] blog archive is broken


flynnduism:
Flagged by @squillace - this url is broken, https://www.helm.sh/blog/page2/
looks like a CDN routing issue

flynnduism:
Closing, resolved via #152

Issue #97, opened 2018-06-01
fix CDN url redirection rules


bacongobbler:
As of right now, https://www.helm.sh/blog/ does not redirect to https://www.helm.sh/blog/index.html. We originally did not require this because helm.sh was a single-page site. Now with the blog up, we should probably add some CDN rewrite rules so that we can tweet out cleaner URL links.

PR #152, opened 2018-11-15
Blog usability & clarifications

bridgetkromhout:
- Found that the helm blog was a bit difficult to read since we weren't using the "read more" function that was configured in the Jekyll installation, making some posts very very long on the main blog page. I added a "read more" break to most posts (except a very short one), designated with a small right arrow and some aria accessibility clarifications.

- Noted that `author` and `authorname` were used inconsistently; I normalized and documented them. I removed a hyphen which ends up trailing awkwardly if someone doesn't use `author`, and added some missing author names.

- Fixed the broken links to Older/Newer posts (which had to do with blobstore not automatically checking `index.html`) and also fixed a couple of typos.


bridgetkromhout:
And yes, I see https://github.com/helm/helm-www/issues/146 and https://github.com/helm/helm-www/issues/97 - adding index.html is a band-aid but probably better than nothing for the moment.

PR #166, opened 2018-12-12
Remove negative margin to prevent footer overlap and double-scrolling

lgfa29:
It's a small detail, but this CSS rule causes the footers to overlap, which results in an annoying double-scrolling in some of the longer pages.

screen shot 2018-12-12 at 12 40 21 pm

screen shot 2018-12-12 at 12 39 36 pm


flynnduism:
Thanks @lgfa29, this looks good. Merging!

PR #169, opened 2019-01-08
update packages for security vulnerabilities

flynnduism:

PR #162, opened 2018-12-07
Merge Sites

flynnduism:
Looking to merge the `helm.sh` Jekyll site and the `docs.helm.sh` Hugo site into a single, Hugo-based app for easier deployment and hosting.

PR #170, opened 2019-01-08
Update packages

flynnduism:
Two security vulnerabilities remained following #169, so this is address those - it's necessary to update both sites from Gulp `3.9.1` to `4.0.0` to get the dependancies in order.

Gulp 4 has some breaking changes though, so here I've had to update both `gulpfile.js` to work with the new version. ☔️


flynnduism:
Closing this as security alerts were resolved via #162

PR #62, opened 2017-11-03
Fix wrong reference to listo in sidebar

neil-ca-moore:
See https://github.com/kubernetes/helm/issues/3090

PR #176, opened 2019-02-07
Fix "helms" showing in in the title and header of the docs

carolynvs:
The hugo configuration settings were _under_ the params section which in TOML means that they are part of params, so hugo was completely ignoring them all.

I have moved them up a bit in the file so that they are parsed as root level values in the file. Now the pluralizeListTitles setting (along with a few others) will be applied by Hugo properly.

### DIFF
--- a/docs.helm.sh/config.toml
+++ b/docs.helm.sh/config.toml
@@ -6,17 +6,16 @@ theme = "helmdocs"
 contentDir = "source/docs"
 publishdir = "app"
 
-[params]
-    title = "Helm Docs"
-    author = "Ronan Flynn-Curran"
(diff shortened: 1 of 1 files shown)


flynnduism:
Thanks @carolynvs  this is a great catch. LGTM

CI for the site is out of action until #62 ships... in the meantime the above checks from Netlify won't work here, they should be ignored.

carolynvs:
Here is what the site looks like locally with a hugo serve
screen shot 2019-02-07 at 11 20 07 am

carolynvs:
oh by the way, I think you'll want to merge this into your open WIP PR. I think I saw the same thing in that config.toml too. 👍

PR #178, opened 2019-02-08
Resolve hugo build error

flynnduism:
This is a follow up to #176 - the `relativeURLs` option in the site config was throwing errors when hugo was building the site. Removing it will fix the builds.


flynnduism:
_Note:_ the CI checks shown here won't pass for this PR, as the current site structure doesn't work with Netlify... the checks from there should be ignored until #162 is resolved.

bacongobbler:
LGTM!

Issue #188, opened 2019-03-14
mobile layout has overlapping content


flynnduism:
On mobile, the logo and main menu are overlapping, resulting in a messy visual and making it hard to click/tap the menu links.

![IMG_5049](https://user-images.githubusercontent.com/686194/54370938-c55e1880-4635-11e9-92cc-02660df035ae.PNG)

Thanks @bridgetkromhout for flagging this!

flynnduism:
Fixed via #189

Issue #85, opened 2018-04-25
Helm.sh website using large amounts of CPU


brendanstennett:
Not related to helm specifically but visiting the website causing my CPU to spike and stay there until closing it. It's not just the home page either, other pages seem to have the same problem.

@bacongobbler suggested in https://github.com/kubernetes/helm/issues/3959 that it might be the animations.  I think he might be correct on that one.

It should be considered a high priority issue though because leaving this page open in any browser tab is going to kill a notebook battery in a hurry.

bridgetkromhout:
Is it possible for the boat to (by default) make a trip or two across the page and then pause? Perhaps to be reactivated by a click? (I like watching the boat, but of course I understand we shouldn't over-use people's local resources.)

flynnduism:
@bridgetkromhout we could add some js to stop/pause the animation like you say. Currently there's no JS in play, it's using CSS for keyframe animations. 

**CSS is ~~awesome~~**
Depending on complexity, CSS animation can be pretty [resource heavy](https://www.youtube.com/watch?v=1ZWugkJV5Ks); in this situation it's largely down to how the animation is written with [lots of incremental steps](https://github.com/helm/helm-www/blob/master/themes/helm/static/src/sass/helm-animation.scss#L8L17) - each cycle is actually plotting 10 sub-animations to simulate waves. Plus it's animating along both X and Y axis so it's taxing the browser to calculate a lot of things.

**Simplify things**  
I'd like to make the boat into a bobbing up/down `.gif`, and have than follow a linear path from left to right. I think this would be a quick way to reduce the amount of keyframe animation required of the browser by 90%.

![boat-sm](https://user-images.githubusercontent.com/686194/54379701-1f1b0e80-4647-11e9-9f21-304180d3181c.gif)

e.g. move this along the X axis, elimate the complicated (and incremental) Y axis keyframes.

---

This has been sitting in the issue queue for awhile as I've been prioritizing other items (broken links, infra migration) over it, I must apologize for the neglect.

I will take a look at this today and see how to simplify the CPU demands.

flynnduism:
Closing this following #189

PR #189, opened 2019-03-14
rework the css boat animation

flynnduism:
Opening this to address #85. This reduces the animation from having 10 keyframes (w/ horizontal and vertical movements) to 2 keyframes of horizontal movement.

This is mostly done by using an animated gif for the boat, reducing the amount of css animation render work asked of the browser, which can be pretty cpu-gobbley.


bacongobbler:
I noticed that the boat accelerates and decelerates due to the key-frame animation changes. Is there a way to make the acceleration at zero (constant horizontal movement)?

flynnduism:
@bacongobbler yep that's the goal here. Simplifying the code = simpler movement.

Preview here: https://5c8ab2a3594fd40008ce5990--helm-merge.netlify.com 
@brendanstennett could you take a look at that url and see if the issue persists?

flynnduism:
tweaked CSS to fix #188 too

Screen Shot 2019-03-14 at 12 45 21 PM

bridgetkromhout:
Both the boat-animation change and the mobile layout content overlap fix == LGTM.

PR #199, opened 2019-04-11
ref: allow displaying multiple versions of the docs

bacongobbler:
With this change, each version of the docs added to the site data can be displayed on the site.

This also makes any requests to /docs redirect to /docs/latest.

TODO:

- [ ] make the nav bar show a dropdown list (cc @flynnduism)
- [ ] tie in with @flynnduism's work to pull multiple versions of the docs into the `content/docs` directory
- [ ] test with netlify to ensure all existing URLs still work


bacongobbler:
closing; will have to revisit this at some point in the future but I don't believe this is the right approach.

PR #222, opened 2019-05-16
fix(go-import): fix go path on meta tags

Inozuma:
Hello, as discussed on #helm-dev slack, this pull request fixes meta tags for go tools.

More info on go-import meta tag: https://golang.org/cmd/go/#hdr-Remote_import_paths
And go-source meta tag: https://github.com/golang/gddo/wiki/Source-Code-Links


bacongobbler:
Thanks for the fix!

Issue #233, opened 2019-07-17
CLI docs - Option descriptions require side scrolling


seandilda:
When reading the command line docs, the element to describe the options is artificially narrow (lots of whitespace to the left and right), and side scrolling is enabled.

This makes the docs very difficult to understand.  To see what's available for a command like 'helm install', a user has to to constantly scroll left and right.  Aside from slowing down reading, this also makes it easy to lose visual indexing between the description and option name as you can only keep one one the screen at a time.

flynnduism:
Fixed and deployed via #235. Closing this issue.

PR #235, opened 2019-07-17
#233 (docs): Improve layout for code line length

flynnduism:
Fix for #233. 

Having horizontal scrolling within code snippets is standard web practice, but I do see how lots of sideways scrolling is annoying when the code runs far longer than the reading pane.

The website layout limits line length intentionally, which imposes whitespace on large screens. This approach is to [aid readability of page copy](https://uxplanet.org/10-tips-on-typography-in-web-design-13a378f4aa0d#f63f), but does not suit the code sections. 

**Before:**

Screen Shot 2019-07-17 at 2 40 53 PM

The proposed solution here widens the code boxes to the right towards the edge of the viewport, to give long lines of code more space. This will maintain a pleasant line-length for the normal text, but expand for the code.

**After:**

Screen Shot 2019-07-17 at 2 31 42 PM

### DIFF
--- a/themes/helm/static/src/sass/docs-content.scss
+++ b/themes/helm/static/src/sass/docs-content.scss
@@ -73,9 +73,11 @@
     background-color: #e4f3f9;
     color: black;
     border: none;
-    padding: 0.75em 1em;
+    padding: 0.75em 2rem 0.75rem 1rem;
     margin: 0.25em 0 2.5em;
     font-size: 1rem;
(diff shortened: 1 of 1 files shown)


flynnduism:
@seandilda here's a preview of this change to the docs. 

Would this approach help your issue?
https://deploy-preview-235--helm-merge.netlify.com/docs/helm/

seandilda:
That looks much better, thank you!

PR #230, opened 2019-07-03
ref(docs): import docs into v3.helm.sh

bacongobbler:
DO NOT MERGE until we get consensus from the core maintainers on this.

This is a significant development change to the documentation. Essentially, this moves the content of the `docs/` dir in helm/helm over to helm-www. This development approach is similar to kubernetes/kubernetes and kubernetes/website, where documentation now is PR'd against the website.

This provides several significant enhancements over the existing system:

1. No more URL redirect munging in gulpfile.js, which has caused us several issues in the past... https://github.com/helm/helm-www/issues/229 https://github.com/helm/helm-www/issues/190 https://github.com/helm/helm-www/issues/185 https://github.com/helm/helm-www/issues/181 https://github.com/helm/helm-www/issues/161 and https://github.com/helm/helm-www/issues/157, to name a few.
2. No need to clone helm/helm to build the site, which is a big blocker for adoption.
3. No need to open multiple PRs when introducing new documentation material (one for the docs, and one to add to the menus).
3. An opportunity to provide alternative locales, such as chinese or spanish.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

PR #244, opened 2019-07-23
[230] ensure gulp does not wipe away content/docs/

flynnduism:
I noticed running gulp will erase the content/docs/ directory during the `clean` task. This is a legacy of the previous cloning of the docs from helm/helm, and should be removed now as part of #230.

This is a short term fix until #237 (Hugo Pipes) is implemented and Gulp is removed.


bacongobbler:
good catch.

PR #245, opened 2019-07-23
[V3][#240] add search to v3 docs

### LINKED ISSUE #240: Search box does not work
Hello, it seems that the top left search field in [this page](https://helm.sh/docs/) does not work (`Search...` field).

flynnduism:
The search box is missing on the v3 docs. This PR reimplements the search in the sidebar.

Closes #240.

### DIFF
--- a/layouts/docs/list.html
+++ b/layouts/docs/list.html
@@ -49,3 +49,5 @@
 
   </div>
 </div>
+
+{{ partial "js.html" . }}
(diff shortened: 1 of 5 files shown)


bacongobbler:
Do we need a separate key for v3.helm.sh? Looking at [agolia's docsearch config for Helm](https://github.com/algolia/docsearch-configs/blob/master/configs/helm.json), it looks like it only crawls helm.sh so it won't be able to crawl v3.helm.sh. Perhaps we should revisit this when we're ready to switch over to Helm 3 as the default for helm.sh?

flynnduism:
Hmm, good catch. Creating seperate keys may well require separate site indexes, to have a specific search for each version to avoid crossed wires? I agree let's close this and revisit when v3 is ready to be switched over.

Kubernetes.io search results are all pointing to the main prod version of the site, regardless of which version of the site you are running the search on. It's confusing IMO, you don't want a search to unexpectedly change versions on you. Removing/disabling search for legacy releases avoids this confusion.

PR #246, opened 2019-07-24
[#230] fix the grid layout for v3 docs

flynnduism:
This PR is the first in a series of layout fixes to restore the Hugo templates for the docs. During the build pipeline changes in #230, there were also html and css changes to the theme files that bumped a few things around.

Live site:

Screen Shot 2019-07-24 at 12 29 50 PM

After this patch:

Screen Shot 2019-07-24 at 12 29 59 PM

---

## To Do:

I will follow up with a separate PR to tidy up the _category_ landing pages (like [this one](https://v3.helm.sh/docs/topics/chart_template_guide/)).

---

Signed-off-by: flynnduism <flynnduism@gmail.com>

### DIFF
--- a/layouts/docs/list.html
+++ b/layouts/docs/list.html
@@ -1,9 +1,9 @@
 {{ partial "header.html" . }}
 
-<div class="container container-full docs-home">
+<div class="container container-full">
   {{ partial "sidebar.html" . }}
 
-  <div class="main home" id="scrollpane">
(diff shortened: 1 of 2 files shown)


bridgetkromhout:
https://deploy-preview-246--helm-v3-dev.netlify.com/docs/faq/ looks much better!

PR #254, opened 2019-08-08
Updating zoom link for weekly call

bridgetkromhout:
Looks like the link for the weekly developer call on the website was an old link. i suspect people were finding the right link from https://github.com/helm/helm/blob/master/README.md#community-discussion-contribution-and-support but if they looked on https://helm.sh it was wrong.


bacongobbler:
thanks!

PR #256, opened 2019-08-12
Change banner to mention beta.1

bridgetkromhout:
Currently, the banner atop the website mentions alpha 1, but we're on alpha 2 now.

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>

PR #269, opened 2019-08-27
fix sticky banner for v2 <-> v3 site links

flynnduism:
* denotes latest releases as `2.14.3` and `3.0.0.beta.1`, this replaces PR #256 which got stuck with a conflict
* restores the sticky v2/v3 banner to the homepage *

The sticky banner was taken off the homepage when the version dropdown menu was added to the [site](https://github.com/helm/helm-www/commit/a392fbc9496223b2cc39c1423c575d179a72d8b3). Adding it back here because I think both are useful to highlight where things are.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -30,7 +30,7 @@ defaultContentLanguage = "en"
 
 [[params.versions]]
 version = "v3.0.0"
-patch = "beta.1"
+patch = "beta.2"
 url = "https://v3.helm.sh/docs"
 primary = true
 
--- a/config.toml
+++ b/config.toml
@@ -28,6 +28,21 @@ defaultContentLanguage = "en"
     "app"
   ]
 
+[[params.versions]]
+version = "v3.0.0"
+patch = "beta.1"
+url = "https://v3.helm.sh/docs"
+primary = true
+
+[[params.versions]]
+version = "v2.14.3"
+patch = "stable"
+url = "https://helm.sh/docs"
+
+[[params.versions]]
+version = "v2.14.0"
+url = "https://v2-14-0.helm.sh/docs"
+
 [Permalinks]
   code = "/:filename/"
 
--- a/themes/helm/layouts/partials/banner.html
+++ b/themes/helm/layouts/partials/banner.html
@@ -2,15 +2,15 @@
   <div class="cc-container">
 
     <!-- viewing helm 2 version of site (desktop) -->
-    <p class="center text-center show-for-medium-up">You are viewing Helm 2 (latest stable). &nbsp; Helm <a href="https://v3.helm.sh"><strong>3.0.0-beta.1 is here</strong>. &nbsp; Visit the <a href="https://v3.helm.sh"><strong>v3 docs</strong></a> or read the <a href="https://helm.sh/blog"><strong>blog</strong></a> for details.</p>
+    <p class="center text-center show-for-medium-up">You are viewing Helm 2 (latest stable). &nbsp; Helm <a href="https://v3.helm.sh"><strong>3.0.0-beta.2 is here</strong>. &nbsp; Visit the <a href="https://v3.helm.sh"><strong>v3 docs</strong></a> or read the <a href="https://helm.sh/blog"><strong>blog</strong></a> for details.</p>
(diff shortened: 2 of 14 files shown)


themes/helm/layouts/partials/nav.html
```diff
@@ -5,10 +5,10 @@
 <li class="social"><a href="https://twitter.com/helmpack" target="_blank" class="hide-for-small-only" title="Helm on Twitter"><img src="/img/twitter.svg" alt="Helm on twitter" /></a></li>
 <li class="social"><a href="https://github.com/helm/helm" target="_blank" class="hide-for-small-only" title="Helm on Github"><img src="/img/github.svg" alt="Github" /></a></li>
 <li class="versioner">
-  <a data-dropdown="drop1" aria-controls="drop1" aria-expanded="false" class="versioner-trigger">v2.14.2 <i class="fa fa-caret-down"></i></a>
+  <a data-dropdown="drop1" aria-controls="drop1" aria-expanded="false" class="versioner-trigger">v2.14.3 <i class="fa fa-caret-down"></i></a>
   <ul id="drop1" class="f-dropdown" data-dropdown-content aria-hidden="true" tabindex="-1">
-    <li><a href="https://v3.helm.sh/docs">v3.0.0 <em>alpha 1</em></a></li>
-    <li><a href="https://helm.sh/docs" class="active">v2.14.2 <em>stable</em></a></li>
+    <li><a href="https://v3.helm.sh/docs">v3.0.0 <em>beta 1</em></a></li>
```
bridgetkromhout: We're now at beta 2 (hooray!)

themes/helm/layouts/partials/banner.html
```diff
@@ -0,0 +1,16 @@
+<div id="banner">
+  <div class="cc-container">
+
+    <!-- viewing helm 2 version of site (desktop) -->
+    <p class="center text-center show-for-medium-up">You are viewing Helm 2 (latest stable). &nbsp; Helm <a href="https://v3.helm.sh"><strong>3.0.0-beta.1 is here</strong>. &nbsp; Visit the <a href="https://v3.helm.sh"><strong>v3 docs</strong></a> or read the <a href="https://helm.sh/blog"><strong>blog</strong></a> for details.</p>
+
+    <!-- viewing helm 2 version of site (mobile) -->
+    <p class="center show-for-small-only"><a href="https://helm.sh/docs"><small style="font-size: 0.9rem;">Viewing Helm 2 (latest). &nbsp; Helm 3.0.0-beta.1 is here - <a href="https://v3.helm.sh" target="_blank"><strong>Docs</strong></a> | <a href="https://helm.sh/blog/" target="_blank"><strong>Blog</strong></a></small></a></p>
```
bridgetkromhout: These two choices mention beta 1, but happily we're now at beta 2.

flynnduism:
Thanks @bridgetkromhout! I've updated the label to `beta.2`, and cherry-picked Luc's work from the v3 docs which parameterizes the versioning stuff.

config.toml
```diff
@@ -28,6 +28,21 @@ defaultContentLanguage = "en"
...
   ]
 
+[[params.versions]]
+version = "v3.0.0"
+patch = "beta.2"
+url = "https://v3.helm.sh/docs"
+primary = true
+
+[[params.versions]]
+version = "v2.14.3"
```
bacongobbler: Would it be best to call this version "latest" instead? That way we don't have to make a new PR to bump it every time a new release is out.

flynnduism: Good idea - that would simplify the site handling of versions. 

_Current live site_:

Screen Shot 2019-09-04 at 9 19 13 AM

_Simpler labels suggestion_:

Screen Shot 2019-09-04 at 9 27 10 AM
 
---

I worry that latest is too vague when we are presenting v2 and v3 choices. The version number is significant as the docs are very different between the two, and I think clarifying whether it's alpha, beta or RC helps show how close v3 is to becoming 'latest'. 

I'd like to keep manually adding as much clarification info as possible for now, and switch to having the top, default choice read `latest` when 3 is released. It'll be much simpler then.

PR #280, opened 2019-09-11
Update helm-home.scss

bsradcliffe:
Improve readability of features


bsradcliffe:
Don't mind me, just a designer walking through with severe leading OCD. Just makes the feature blocks a bit more compact and easier to scan.

PR #305, opened 2019-10-09
Add vanity url for helm v3 with Go modules

### LINKED ISSUE #300: v3 vanity url for imports
`import helm.sh/helm` worked until we moved to Go modules and v3. Now we need to have `import helm.sh/helm/v3`. There needs to be an additional redirect put in place for this.

You can read the docs on this stuff at https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher.

mattfarina:
With Go modules and packages with a major version greater than 1
the major version needs to be in the path. In Helm's case this
is a sort of virtual path and not on disk. The vanity redirect
needs to reflect this new version.

Fixes #300

### DIFF
--- a/content/code/helm-v3.md
+++ b/content/code/helm-v3.md
@@ -1,0 +1,6 @@
+---
+url: "helm/v3"
+name: "helm/v3"
+repoURL: "https://github.com/helm/helm"
+branch: "dev-v3"
+---


rimusz:
LGTM

mattfarina:
Note, this is not ready to go. The redirect isn't actually working.

mattfarina:
PR appears to be good to go now

povils:
When it is expected to be enabled?:)

bacongobbler:
As soon as a PR is merged to master, the site is rebuilt. It's available now!

It will redirect straight to helm/helm, but the URL is now available for Go mods to pick up.

https://helm.sh/helm/v3

PR #306, opened 2019-10-09
fix: Update docs to include generate-name argument

mcristina422:
install: requires release name or --generate-name argument
`helm install [NAME] [CHART] [flags]`


bacongobbler:
thanks for catching this. LGTM.

PR #307, opened 2019-10-14
Fix broken link in Quicklinks

sco11morgan:
Looks like this change didn't get migrated over.

https://github.com/helm/helm/pull/6067


bacongobbler:
hi @sco11morgan, Helm requires that you agree and sign off on the DCO before we can merge your PR. Would you mind fixing that up? Thanks!

PR #309, opened 2019-10-16
feat(rbac): add Role-based Access Control documentation

bacongobbler:
closes https://github.com/helm/helm/issues/5613

### DIFF
--- a/content/docs/intro/rbac.md
+++ b/content/docs/intro/rbac.md
@@ -1,0 +1,147 @@
+---
+title: "Role-based Access Control"
+description: "Explains how Helm interacts with Kubernetes' Role-Based Access Control."
+weight: 3
+---
+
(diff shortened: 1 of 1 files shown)


content/docs/intro/rbac.md
```diff
@@ -0,0 +1,147 @@
...
+
+$ kubectl create rolebinding sam-secret-reader
+    --clusterrole secret-reader \​
+    --user sam \​
+    --namespace foo
+```
+
+### Example: Grant a user read-only access at the cluster scope
+
+In certain scenarios, it may be beneficial to grant a user cluster-scope access. For example, if a user wants to run the command `helm list --all-namespaces`, the API requires the user has cluster-scope read access.
```
pc-mreeves, line 131: @bacongobbler - Am I the only one that thinks this is insane?

If I want to allow more users to see metadata about helm deployments via `helm ls -A` I need to grant them access to every secret in the entire cluster?

Would it not be possible to store metadata about deploys into a specific "helm-deploys" namespace and grant access to THAT instead? 

Am I missing something? I understand there may be sensitive info in the helm deploys and that is the reasoning to keep it all secret but this makes no sense to me.  Allowing certain people to possibly see leaked secrets vs. explicitly letting them see everything is worlds apart.

PR #323, opened 2019-10-23
Updates FAQ and Glossary pages with the most recent information

thomastaylor312:

PR #324, opened 2019-10-23
Formats all docs to use same wrapping

thomastaylor312:
There was a wide variety of wrapping formats from 80 chars all the
way up to never wrapping. This changes the wrapping of all documents
to be at/around 80 chars to make it easier to review and read in text
editors.

I made no other changes to content save for sorting the list of Kubernetes
distros alphabetically to avoid anything that looks like preference for
any given distro.

### DIFF
--- a/content/docs/_index.md
+++ b/content/docs/_index.md
@@ -5,9 +5,16 @@ description: "Everything you need to know about how the documentation is organiz
 
 # How the documentation is organized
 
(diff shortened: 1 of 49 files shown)


thomastaylor312:
This is branched off of #323 so as to avoid any merge conflicts. Once that is merged, I'll rebase

thomastaylor312:
Ok, this is rebased and ready for review

PR #375, opened 2019-11-13
[v3] fix /blog redirect loop

flynnduism:
Following the site version changes, the `/blog` pages are inaccessible due to a redirect loop caused by [this code](https://github.com/helm/helm-www/blob/master/netlify.toml#L23-L57).


flynnduism:
Rush merging this to fix a bug on the live site that renders the `/blog` pages inaccessible.

PR #380, opened 2019-11-13
Adding back go get metadata template

mattfarina:


bacongobbler:
Thanks for catching this.

PR #386, opened 2019-11-14
Fix helm inspect format

oneoneonepig:
It should be "helm inspect chart stable/mysql" instead of "helm inspect stable/mysql"

Signed-off-by: Jeffrey Chu <peihuachu1112@gmail.com>

### DIFF
--- a/content/docs/intro/quickstart.md
+++ b/content/docs/intro/quickstart.md
@@ -68,8 +68,11 @@ Released smiling-penguin
 ```
 
 In the example above, the `stable/mysql` chart was released, and the name of our
-new release is `smiling-penguin`. You get a simple idea of the features of this
-MySQL chart by running `helm inspect stable/mysql`.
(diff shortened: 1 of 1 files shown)


content/docs/intro/quickstart.md
```diff
@@ -69,7 +69,7 @@ Released smiling-penguin
 
 In the example above, the `stable/mysql` chart was released, and the name of our
 new release is `smiling-penguin`. You get a simple idea of the features of this
-MySQL chart by running `helm inspect stable/mysql`.
+MySQL chart by running `helm inspect chart stable/mysql`.
```
marckhouzam: Good catch. Might as well put the proper helm v3 format which is `helm show all stable/mysql`

oneoneonepig: I've updated to `helm show chart` and added a short description for `helm show all`

PR #410, opened 2019-11-26
fix some wrong commands

knightXun:
Those `helm install` commands can mislead users, So I fix these, Make it more helpful for greenhand


bacongobbler:
Can you please look into the failing check? Thanks

knightXun:
ping @bacongobbler

PR #419, opened 2019-12-02
Multiple service accounts example correction

naseemkullah:
It makes more sense to have a component's service account values under itself in the values file.


scottrigby:
For more context see [this slack chat](https://kubernetes.slack.com/archives/C6E3XH1ED/p1575306186335000). Helm itself doesn't really care, but this feels better organized and I think this makes more sense as an idiomatic pattern 👍

naseemkullah:
FWIW Chart maintainers tend to follow what's recommended in docs see https://github.com/helm/charts/pull/15108/commits/fb81e4587081821dc3e4f9b733c4f12a79c12b72

PR #430, opened 2019-12-10
[docs] Fix chart repository doc image paths

Sinclert:
The images for supporting the [Chart Repository guide](https://helm.sh/docs/topics/chart_repository/) are not correctly shown.

It may be due to a previous refactor, where the `images` folder used to be at the same level of the `chart_repository.md` documentation file.


Sinclert:
Hi Ronan,

I have applied the changes you told me, using absolute paths instead of relative ones.

PR #448, opened 2019-12-28
Dependency repo alias explained

carlosjgp:
Added all the fields to the `dependencies` item of the `Chart.yaml` file
and an example using repository alias instead the repository URL

Signed-off-by: Carlos Juan Gómez Peñalver <carlosjuangp@gmail.com>

### DIFF
--- a/content/docs/topics/charts.md
+++ b/content/docs/topics/charts.md
@@ -59,7 +59,14 @@ sources:
 dependencies: # A list of the chart requirements (optional)
   - name: The name of the chart (nginx)
     version: The version of the chart ("1.2.3")
-    repository: The repository URL ("https://example.com/charts")
+    repository: The repository URL ("https://example.com/charts") or alias ("@repo-name")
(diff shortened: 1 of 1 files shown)


content/docs/topics/charts.md
```diff
@@ -59,7 +59,14 @@ sources:
...
   - name: The name of the chart (nginx)
     version: The version of the chart ("1.2.3")
-    repository: The repository URL ("https://example.com/charts")
+    repository: The repository URL ("https://example.com/charts") or alias ("@repo-name")
+    condition: (optional) A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )
+    tags: # (optional)
+      - Tags can be used to group charts for enabling/disabling together
+    enabled: (optional) Enabled bool determines if chart should be loaded
+    import-values: # (optional)
+      - ImportValues holds the mapping of source values to parent key to be imported. Each item can be a string or pair of child/parent sublist items.
```
thomastaylor312: It might be helpful to add an example of this, but that is non-blocking in this case

PR #470, opened 2020-01-17
fix(community): Port fixes from v2 release checklist

hickeyma:
Release checklist was updated with some fixes in v2 stream. This commit pulls over those fixes.


hickeyma:
Thanks for review @bacongobbler

PR #463, opened 2020-01-14
Update values_files.md

EverOctivian:
in helm3 you must specify a release name when installing a chart

PR #478, opened 2020-01-27
fix helm install command

bacongobbler:
follow-up from #463.


bridgetkromhout:
Looks like this was solved in https://github.com/helm/helm-www/pull/510 - thanks!

PR #481, opened 2020-02-03
feat(topics): Helm Version Support Policy

bacongobbler:
closes https://github.com/helm/helm/issues/6981

### DIFF
--- a/content/en/docs/topics/version_skew.md
+++ b/content/en/docs/topics/version_skew.md
@@ -1,0 +1,58 @@
+---
+title: "Helm Version Support Policy"
+description: "Describes Helm's patch release policy as well as the maximum version skew supported between Helm and Kubernetes."
+---
+
+This document describes the maximum version skew supported between Helm and Kubernetes.
+
+## Supported Versions
+
+Helm versions are expressed as `x.y.z`, where `x` is the major version, `y` is the minor version, and z is the patch
+version, following [Semantic Versioning](https://semver.org/spec/v2.0.0.html) terminology.
+
+The Helm project maintains a release branch for the most recent minor release. Applicable fixes, including security
+fixes, are cherry-picked into the release branch, depending on severity and feasibility. Patch releases are cut from
+that branch as needed. This decision is owned by the release maintainer.
+
+## Supported Version Skew
+
+When a new version of Helm is released, it is compiled against a particular minor version of Kubernetes. For example,
+Helm 3.0.0 interacts with Kubernetes using the Kubernetes 1.16.2 client, so it is compatible with Kubernetes 1.16.
+
+As of Helm 3, Helm is assumed to be compatible with `n-3` versions of Kubernetes it was compiled against. Due to
+Kubernetes' changes between minor versions, Helm 2's support policy is slightly stricter, assuming to be compatible
+with `n-1` versions of Kubernetes.
+
+For example, if you are using a version of Helm 3 that was compiled against the Kubernetes 1.17 client APIs, then it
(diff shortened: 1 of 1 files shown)


content/en/docs/topics/version_skew.md
```diff
@@ -0,0 +1,59 @@
...
+It is not recommended to use Helm with a version of Kubernetes that is newer than the version it was compiled against,
+as Helm does not make any forward compatiblility guarantees.
+
+If you choose to use Helm with a version of Kubernetes that it does not support, you are using it at your own risk.
+
+Please refer to the table below to determine what version of Helm is compatible with your cluster.
+
++--------------+-------------------------------+
+| Helm Version | Supported Kubernetes Versions |
+|--------------|-------------------------------+
```
mattfarina: The table is not showing up correctly in the preview. See https://deploy-preview-481--helm-merge.netlify.com/docs/topics/version_skew/

Ways to fix:

* Remove the first and last lines of the table. The current line 38 should be the first.
* On the separator between the table header and body the last character should be changed from `+` to `|`.

bacongobbler: thanks for catching this. So many different ways to represent a table!

willysamz:
is the content of this table accurate?

tejaswiniVadlamudi:
It's a very nice idea to show compatibility across different k8s and helm versions. Not sure if my questions are supposed to be addressed here.

- Is the original idea to help helm users to select a proper helm version? 

- In general, how do we expect helm users to select a minimum helm version?

bacongobbler:
The idea here is to provide users with information with regards to what versions of Helm is compatible with their Kubernetes cluster, so in effect, yes, it is to help users select the correct version of Helm.

>     * In general, how do we expect helm users to select a minimum helm version?

I'd highly recommend choosing a Kubernetes release that matches the same version Helm was compiled against. For example, Helm 3.0.0 interacts with Kubernetes using the Kubernetes 1.16.2 client, so I'd try and match it with Kubernetes 1.16 for the best experience. We've found that matching or going one version below usually results in fewer bugs, as Kubernetes does break compatibility from time to time and it may take us a release or two to find regressions.

bacongobbler:
fixed table formattting and added Helm 3.2 to the matrix

bacongobbler:
@mattfarina do the changes here look good to you? Looking to bring this in so we have a place to point community members at for our support matrix. Thanks! :)

content/en/docs/topics/version_skew.md
```diff
@@ -0,0 +1,57 @@
...
+fixes, are cherry-picked into the release branch, depending on severity and feasibility. Patch releases are cut from
+that branch as needed. This decision is owned by the release maintainer.
+
+## Supported Version Skew
+
+When a new version of Helm is released, it is compiled against a particular minor version of Kubernetes. For example,
+Helm 3.0.0 interacts with Kubernetes using the Kubernetes 1.16.2 client, so it is compatible with Kubernetes 1.16.
+
+As of Helm 3, Helm is assumed to be compatible with `n-3` versions of Kubernetes it was compiled against. Due to
+Kubernetes' volitility as a newer project, Helm 2's support policy is slightly stricter, assuming to be compatible
```
hickeyma, line 23: I think this would read better as:

"Due to Kubernetes' volitility as a newer project" --> "Due to Kubernetes' changes between minor versions"

Vaccano:
This is the link for the docs (for reference):

https://helm.sh/docs/topics/version_skew/#:~:text=As%20of%20Helm%203%2C%20Helm,n%2D1%20versions%20of%20Kubernetes

PR #493, opened 2020-02-13
Helm 3.1.0 is out. Updating the listed version

mattfarina:


flynnduism:
> Thanks for updating @mattfarina. Should this be removed from the file?

We can remove/include any other releases in this dropdown menu - what major releases would be the most useful to preserve archival docs for? IMO any version that has a considerable usage base should be maintained and navigable like this - but not sure which releases best warrant that.

bacongobbler:
As long as the versioning system continues to work, I think it's worth keeping them around for as long as possible. I love reviewing super-old versions of the documentation for historical purposes.

On older releases of Django ([like 2.0](https://docs.djangoproject.com/en/2.0/)), there's a banner mentioning that they're viewing a version of the docs that is unsupported. I think that strikes a great middleground: users can still view older "archived" releases, but it's clear that it's not something we actively support.

PR #496, opened 2020-02-14
Adding a step to the release checklist to update the Helm website

mattfarina:


content/en/docs/community/release_checklist.md
```diff
@@ -372,7 +372,11 @@ always beneficial as it can be easy to miss something.
 
 When you are ready to go, hit `publish`.
 
-## 9. Evangelize
+## 9. Update Docs Site
+
+The [Helm website docs section](https://helm.sh/docs) lists the Helm versions for the docs. The minor and patch information needs to be updated on the site. To do that create a pull request against the [helm-www repository](https://github.com/helm/helm-www). In the `config.toml` file find the proper `params.versions` section and updated the Helm version.
```
hickeyma: A few small things but might not be that obvious if you haven't done the release before:
- Is it worth saying that `config.toml` is in the root of the repo?
- There are 3 `params.versions` sections in the doc! :) No real definition what 'proper` is?

mattfarina: When someone does the release their first time I would expect them to pair with another maintainer who can show them. Even when they ride along and do it themselves. That is to do good mentoring.

What versions should be listed? I filed #497 because I had the same question. It's now on the agenda for this weeks helm meeting.

PR #502, opened 2020-02-21
Fixed uninstall example output screen

TribalNightOwl:
Signed-off-by: Hugo Gayosso <tribrhy@gmail.com>


TribalNightOwl:
@bacongobbler , you are right, sorry for the noise, I will close the PR.

I think my brain was expecting to see something more explicit, either an "uninstalled" or a "before/after" screenshot.

PR #513, opened 2020-02-27
helm2 command changed as helm3 versioned

developer-guy:


developer-guy:
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/ 

inside this page theres is also old helm2 command , 

`Now re-running helm install --dry-run --debug ./mychart will produce this YAML:`

no name provided again , this will get an error so i fixed that.

PR #518, opened 2020-02-28
match real output

NargiT:
I was doing the "getting started" and I notice that the documentation does not give the same output as my cli.


content/en/docs/chart_template_guide/getting_started.md
```diff
@@ -200,6 +202,8 @@ NAME: clunky-serval
 LAST DEPLOYED: Tue Nov  1 17:45:37 2016
 NAMESPACE: default
 STATUS: DEPLOYED
+REVISION: 1
+TEST SUITE: None
```
hickeyma: Comment above applies here also.

content/en/docs/chart_template_guide/getting_started.md
```diff
@@ -117,6 +117,8 @@ NAME: full-coral
 LAST DEPLOYED: Tue Nov  1 17:36:01 2016
 NAMESPACE: default
 STATUS: DEPLOYED
+REVISION: 1
+TEST SUITE: None
```
hickeyma: Resources are no longer output in Helm 3.  Output is no as follows:
```console
$ helm install full-coral ./mychart
NAME: full-coral
LAST DEPLOYED: Mon Mar  2 10:09:29 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None

```

The following sentence is no longer valid : `In the output above, we can see that our ConfigMap was created. `

content/en/docs/chart_template_guide/getting_started.md
```diff
@@ -117,6 +117,8 @@ NAME: full-coral
...
 
 RESOURCES:
```
NargiT: Ok I will remove this part also. I though it was only there to display what happends on the cluster and was not related to helm output. Will do this tonight. Thx for the feedback

hickeyma: Great, thanks

PR #519, opened 2020-03-02
Add note about tested release

hickeyma:
This note was recommended when discussing the addition of tests for s390x platform (https://github.com/helm/helm/pull/7096).

It was decided at the meeting that Helm community would continue to accommodate builds  for different platforms and allow tests for those platforms but would be performing automated tests for the Linux AMD64 platform only (as previously). The responsibility however for the testing and validation of Helm for the other platforms is the responsibility of the community which requests Helm for that platform.


hickeyma:
Thanks for the reviews @bacongobbler @mattfarina

PR #532, opened 2020-03-10
Updating the developer and source docs

mattfarina:
* Removing the listing for Dep as it is no longer used to manage
  dependencies
* Updated the compile from source directions as the GOPATH is
  no longer needed to install Helm


mattfarina:
@bacongobbler thanks for the review.

PR #545, opened 2020-03-20
feat(chart_best_practices): Add details about --dry-run with CRDs

hickeyma:
Add explanation about --dry-run flag not supported when CRDs in a chart.

Related: https://github.com/helm/helm/issues/7449


hickeyma:
Thanks for the review @technosophos. The explanation should be good as it was mostly your words! 😃

PR #551, opened 2020-03-30
add missed field of built-in object in doc

theshamuel:


theshamuel:
> @theshamuel Thanks for pushing this PR. `.Release.Time` is no longer supported in Helm v3. Have a look at the thread in [helm/helm#5732](https://github.com/helm/helm/issues/5732) for more details.

thanks for fast reply, will close then this PR

PR #556, opened 2020-04-06
Stubbing out Korean for translation

mattfarina:


flynnduism:
Excited to see this. You've added the elements needed to render Korean content on the site. 👌 

Will content be added to this _or_ a future PR? I ask because the `nav.html` files needs an edit ([here](https://github.com/helm/helm-www/blob/master/themes/helm/layouts/partials/nav.html#L32L40)) in order to display the language option in the site menu, once translated content exists.

mattfarina:
@flynnduism content will come in a future commit. I also created #557. We have folks at Samsung who are doing translations. I am trying to set them up to be able to do it for Helm if them choose.

config.toml
```diff
@@ -111,6 +111,38 @@ name = "Community"
...
+# name = "Charts"
+# url = "https://hub.helm.sh/"
+# weight = 3
+# external = "yes"
+
+#[[languages.zko.menus.main]]
+#name = "Blog"
+#url = "/blog"
+#weight = 2
+
```
hickeyma, line 142: Should community be added for Korean and Chinese also?
```
[[languages.en.menus.main]]
name = "Community"
[...]
```

config.toml
```diff
@@ -111,6 +111,38 @@ name = "Community"
...
+# url = "/docs"
+# weight = 2
+
+# [[languages.ko.menus.main]]
+# name = "Charts"
+# url = "https://hub.helm.sh/"
+# weight = 3
+# external = "yes"
+
+#[[languages.zko.menus.main]]
```
hickeyma, line 138: zko --> ko

config.toml
```diff
@@ -111,6 +111,38 @@ name = "Community"
...
+#name = "Blog"
+#url = "/blog"
+#weight = 2
+
+# [languages.ko.params]
+# language_alternatives = ["en"]
+
 # Chinese
 # [languages.zh]
```
hickeyma, line 147: Just wondering can `zh` be used standalone or does it need to be specific e.g. `zh-cn`?

content/ko/docs/_index.md
```diff
@@ -0,0 +1,6 @@
+---
+title: "Docs Home"
+description: "Everything you need to know about how the documentation is organized."
+---
+
+Translate this doc into Chinese, please!
```
hickeyma, line 6: this doc into Chinese --> this doc into Korean ??

i18n/ko.toml
```diff
@@ -0,0 +1,204 @@
...
+
+[community_contrib_subtitle]
+other = "Helm always welcomes new contributions to the project!"
+
+
+[community_contrib_desc]
+other = """
+### Where to begin?
+
+Helm is a big project with a lot of users and contributors. It can be a lot to take in!s
```
hickeyma, line 189: take in!s --> take in!

i18n/ko.toml
```diff
@@ -0,0 +1,204 @@
...
+other = "Events"
+
+[community_upcoming_events_title]
+other = "Upcoming Events"
+
+[community_upcoming_events_desc]
+other = "Watch this space..."
+
+[community_past_events_title]
+other = "Upcoming Events"
```
hickeyma, line 121: Upcoming --> Past

config.toml
mattfarina: Good idea. done.

config.toml
mattfarina: Thanks for catching that. Fixed.

config.toml
mattfarina: That's a great question. I dug into that when I wrote some docs on localization recently. You can find those at https://helm.sh/docs/community/localization/. Basically, we use the two letter language code in the current setup. Not sure what that does for things like the color vs colour debate.

content/ko/docs/_index.md
mattfarina: oops. thanks for catching that. You also really noticed my copy/paste strategy to this PR.

i18n/ko.toml
mattfarina: Fixed. This was in the Chinese version and I fixed it there, too. Nice catch.

i18n/ko.toml
mattfarina: This was in the Chinese one as well. Fixed in both.

config.toml
hickeyma: Might be worth investigating further to be sure.  It is a bit since I did localization but there is something about languages with colloquial derivations.  I think it might default `zh` to `zh-cn` in most code languages.

mattfarina:
Thanks y'all.

PR #509, opened 2020-02-25
added lookup function docs

raffaelespazzoli:
Signed-off-by: Raffaele Spazzoli <raffaele.spazzoli@gmail.com>

@bacongobbler @mattfarina here is my first draft. Something is wrong with the site generation, but not being familiar with the process, I can't figure out the issue.
Please help me solve that issue as well as provide feedback to the doc.
Also please suggest any place in the doc where we should reference this page of the documentation

PR #564, opened 2020-04-13
feat(docs): document `lookup`

bacongobbler:
supersedes #509


content/en/docs/chart_template_guide/functions_and_pipelines.md
```diff
@@ -190,6 +190,72 @@ logic that is a little more sophisticated than just inserting a string. In the
 next section we will look at the control structures provided by the template
 language.
 
+## Using the `lookup` function
+
+The `lookup` function can be used to resources from resources in a running cluster. The synopsis of
```
mattfarina, line 195: > The `lookup` function can be used to resources from resources in a running cluster.

Doesn't make sense. I think it means to say...

> The `lookup` function can be used to _lookup_ resources in a running cluster.

Though, some other variation may be what was meant or better.

mattfarina:
One other thing I noticed. The base of the branch, prior to the commit for this, is from before we started using the new theme. It's a little dated (a month old).

bacongobbler:
> The base of the branch, prior to the commit for this, is from before we started using the new theme. It's a little dated (a month old).

Good call! Let me rebase

PR #570, opened 2020-04-13
TranslationKor1

singlemocha:
Signed-off-by: MJ Lee <singlemocha@hotmail.com>


mattfarina:
For those who are wondering... I checked the translations using a translation website. It's also worth noting that singlemocha is fluent in both English and Korean and has worked as a translator. I've found that manual translations by those who know Korean are different (better?) than automation.

PR #579, opened 2020-04-17
fix(charts): Update chart type section for better clarity

hickeyma:
Refactor the chart type section to improve explanation of library chart.

### DIFF
--- a/content/en/docs/topics/charts.md
+++ b/content/en/docs/topics/charts.md
@@ -181,12 +181,12 @@ followed by the [kubernetes/charts](https://github.com/helm/charts) project is:
 ### Chart Types
 
 The `type` field defines the type of chart. There are two types: `application`
-and `library`.  Application is the default type and it is the standard chart
-which can be operated on fully. The [library or helper
-chart](https://github.com/helm/charts/tree/master/incubator/common) provides
+and `library`. Application is the default type and it is the standard chart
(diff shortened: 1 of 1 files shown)


content/en/docs/topics/charts.md
```diff
@@ -181,12 +181,12 @@ followed by the [kubernetes/charts](https://github.com/helm/charts) project is:
...
-which can be operated on fully. The [library or helper
-chart](https://github.com/helm/charts/tree/master/incubator/common) provides
+and `library`. Application is the default type and it is the standard chart
+which can be operated on fully.
+The [library chart]({{< ref "/docs/topics/library_charts.md" >}}) provides
 utilities or functions for the chart builder. A library chart differs from an
-application chart because it has no resource object and is therefore not
-installable.
+application chart because it is not installable and usually doesn't contain any
+resource objects.
```
mattfarina, line 189: > usually doesn't contain any resource objects.

If resource objects exist, what form would they be in where they could be usable?

hickeyma: Well, if templates with resources objects existed in the chart then this would be installable in Helm v2 (no type field). With the type field in v3, we don't install a chart with type 'library` but if changed to `application` then it would be installable.

hickeyma:
Thanks for the review @mattfarina

PR #597, opened 2020-04-26
Add helm-secrets

jkroepke:
Hi,

as in https://github.com/helm/helm-www/pull/567 mentioned correctly the helm-secrets project gets inactive.

helm-secrets is a great plugin because with you want to store secret in an encrypted format. encryption on rest is default compliance rule in a lot of companies in germany.

Since is this mandatory for us I will continue to maintain it.

URL to project: https://github.com/jkroepke/helm-secrets

### DIFF
--- a/content/en/docs/community/related.md
+++ b/content/en/docs/community/related.md
@@ -29,6 +29,7 @@ request](https://github.com/helm/helm-www/pulls).
   [private] chart repository
 - [helm-schema-gen](https://github.com/karuppiah7890/helm-schema-gen) - Helm Plugin that generates
   values yaml schema for your Helm 3 charts
+- [helm-secrets](https://github.com/jkroepke/helm-secrets) - Plugin to manage and store secrets safely (based on [sops](https://github.com/mozilla/sops)) 
 
 We also encourage GitHub authors to use the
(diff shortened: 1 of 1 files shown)


jkroepke:
> Is this still deprecated? 

It was never deprecated but I didn't see any activity at least for one year. Now it looks zendesk gets the ownership of the repo but there are no public announcements about this nor any other activity on issues or pull requests.

We could list both projects like the [community exporter list on the prometheus wiki](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) does it.

> With the two versions diverge from each other?

Yep. See: https://github.com/jkroepke/helm-secrets/pull/1

The concept is still the same (shell based and using sops as engine) but I change a lot of implantation details.

jkroepke:
Hi @mattfarina ,

I add recently vault support which means the two versions are now diverge from each other. jkroepke/helm-secrets resolved a lot of opened issues that are reported at the zendesk repositroy.

I would propose to add jkroepke/helm-secrets independently from current state of zendesk/helm-secrets.

The status of the maintenance is still unknown for me. A lot of issues or pull requests are not reviewed or accepted for months while a new maintainer committed to just update the copyright files.

I personally declare the repository as abandoned or inactive. That's the reason to fork the repos and I want to announce this to the community of helm.

References: https://github.com/zendesk/helm-secrets/issues/100

PR #600, opened 2020-04-27
Sort the contributor list in alphabetical order

hickeyma:


bridgetkromhout:
This seems reasonable to me; let's check in with @flynnduism to see if there is anything else in the theme that would be affected/need to change and/or if there is a non-obvious need for the current order.

hickeyma:
Thanks @bridgetkromhout @flynnduism for reviews and merge.

Issue #433, opened 2019-12-11
Does Helm v3 require an explicit list of k8s distros it works with?


tomkivlin:
Or should it just state that any Kubernetes distro that is listed as a [Certified Kubernetes Distribution](https://www.cncf.io/certification/software-conformance/#logos) will work?  
Unless there are additional constraints that need to be in place?  In which case, can we detail those, so new additions to the page can prove their compliance for Helm users.

Affected page: https://github.com/helm/helm-www/blob/master/content/docs/topics/kubernetes_distros.md

I'm happy to raise a PR to delete the page, if that's the approach needed.  Briefly discussed this with a couple of the maintainers at Kubecon NA (can't remember names, sorry!).

bacongobbler:
In general, yes, Helm will work with any certified kubernetes distribution. The page was originally put in place to describe any additional setup that may be required to use Helm with those distributions, or other distributions that have not gone through the CKD process (like k3s, for example).

mattfarina:
I think it would be useful to document, likely at the top of this document, that Helm should work with any conformant version of Kubernetes (whether certified or not).

bridgetkromhout:
I've edited this document based on @mattfarina and @bacongobbler's clarifications. Thanks, @tomkivlin!

PR #604, opened 2020-04-28
Adding clarification about distros that work

bridgetkromhout:
In reviewing https://github.com/helm/helm-www/issues/433, I think this could be a reasonable revision of that page. (The page does not have every link or update that would be possible, but this clarification at the top should help with any ambiguity.) Opinions and/or edits from @bacongobbler and @mattfarina would be helpful, as they commented on the original issue. Thanks!

### DIFF
--- a/content/en/docs/topics/kubernetes_distros.md
+++ b/content/en/docs/topics/kubernetes_distros.md
@@ -12,7 +12,6 @@ Kubernetes](https://github.com/cncf/k8s-conformance) (whether
 This document captures information about using Helm in specific Kubernetes
 environments. Please contribute more details about any distros (sorted
 alphabetically) if desired.
-preference.
 
 
 ## AKS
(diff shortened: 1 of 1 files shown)


content/en/docs/topics/kubernetes_distros.md
```diff
@@ -5,12 +5,15 @@ aliases: ["/docs/kubernetes_distros/"]
...
 
+Helm should work with any [conformant version of
+Kubernetes](https://github.com/cncf/k8s-conformance) (whether
+[certified](https://www.cncf.io/certification/software-conformance/) or not).
+
 This document captures information about using Helm in specific Kubernetes
-environments. Distros are sorted alphabetically so as to not imply any sort of
+environments. Please contribute more details about any distros (sorted
+alphabetically) if desired.
 preference.
```
hickeyma, line 15: This should be removed.

mattfarina: Looks like a hanging word was left over...

> preference.

Otherwise the edits look good.

PR #617, opened 2020-05-01
Clarify risks of using configmap backend

jpetazzo:
The main issue when storing Helm release information in configmaps,
is that configmaps are typically broadly readable, while secrets are not.
I tried to add language to clarify that. I deliberately removed the mention
to base64, since a candid reader might have incorrectly assumed that
base64 gives a layer of protection.

Feel free to adapt/reject if this doesn't reflect the initial intent of the doc <3


bacongobbler:
thank you for the expanded notes. LGTM!

PR #623, opened 2020-05-05
add documentation about OCI dependencies

phroggyy:
This adds documentation for the feature added in https://github.com/helm/helm/pull/7613

### DIFF
--- a/content/en/docs/chart_best_practices/dependencies.md
+++ b/content/en/docs/chart_best_practices/dependencies.md
@@ -21,7 +21,7 @@ This will match version `1.2.3` and any patches to that release.  In other
 words, `~1.2.3` is equivalent to `>= 1.2.3, < 1.3.0`
 
 For the complete version matching syntax, please see the [semver
-documentation](https://github.com/Masterminds/semver#checking-version-constraints)
+documentation](https://github.com/Masterminds/semver#checking-version-constraints).
 
 ### Repository URLs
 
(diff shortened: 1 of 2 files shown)


content/en/docs/chart_best_practices/dependencies.md
```diff
@@ -35,6 +35,19 @@ File URLs (`file://...`) are considered a "special case" for charts that are
 assembled by a fixed deployment pipeline. Charts that use `file://` are not
 allowed in the official Helm repository.
 
+#### Experimental support for Charts hosted on OCI registries
+
+If you have [enabled experimental OCI support](/docs/registries/), you can specify
+an OCI reference (`oci://registry/group/image:tag`) for the repository URL.
```
mattfarina, line 41: The language through this and the linked doc make a lot of assumptions about what the reader knows and how they treat helm charts as an image in the registries. I'm not sure if it should be in this PR but we need something in a follow-up that helps someone who isn't an insider understand this.

phroggyy: Yep, this is a good point. I do agree this could be more clear, especially with this being under `chart_best_practices`. I'm hesitant to add details about what an OCI registry _is_ here, that's out of scope. Happy to discuss approaches to this 🤔

jdolitsky: I think this change is good, if anything we need more background in registries page (i can add)

jdolitsky: opened https://github.com/helm/helm-www/issues/637

phroggyy:
@jdolitsky did we want to merge this before the PR itself is merged/released?

jdolitsky:
This or #637 ? I think this is all we need

jdolitsky:
@phroggyy

phroggyy:
Sorry, to clarify, I meant to emphasise _before_. As in before the feature itself (https://github.com/helm/helm/pull/7613) is merged.

jdolitsky:
@phroggyy - ah i see. Yes I jumped the gun on this.. I'll bring this up in the call today, since the docs would go live even before 3.3 released. I can fix if need be

PR #638, opened 2020-05-15
Add a new line to enable code formatting.

ksimka:


bacongobbler:
The documentation listed here is auto-generates from the helm CLI. Please make a PR to update the helm CLI’s help text in helm/helm. Once that is done we can update the documentation in the next release. Thanks!

PR #685, opened 2020-06-10
Update charts_tips_and_tricks.md

sishbi:
Fix issue with imagePullSecret template that was missing the username, password and email within the 'auths' object.
I think email is optional but including for completeness.
Note: this was required when using containerd CRI (such as with MicroK8s)

### DIFF
--- a/content/en/docs/howto/charts_tips_and_tricks.md
+++ b/content/en/docs/howto/charts_tips_and_tricks.md
@@ -164,12 +164,15 @@ imageCredentials:
   registry: quay.io
   username: someone
   password: sillyness
+  email: someone@host.com
 ```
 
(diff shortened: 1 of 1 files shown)


mattfarina:
@sishbi The checks tab at the top has direction on how to correct your commits with a signoff. It has copy and paste commands you can use.

bacongobbler:
You’ll need to rebase and sign off on the first commit. All commits need to be signed off. Instructions are in the failing github check.

sishbi:
@mattfarina @bacongobbler Thanks for the tips. I've never needed to 'sign-off' my commits before. I was trying to do it all within the GitHub website, but I guess that isn't supported.
I've now done as requested and it looks like it was successful.

PR #700, opened 2020-06-16
Adding documentation on template functions

mattfarina:
This includes documentation of most of the template functions
available within Helm templates.

### DIFF
--- a/content/en/docs/chart_template_guide/function_list.md
+++ b/content/en/docs/chart_template_guide/function_list.md
@@ -1,0 +1,1951 @@
+---
+title: "Template Function List"
+description: "A list of template functions available in Helm"
+weight: 6
+---
+
+Helm includes many template functions you can take advantage of in templates. They are listed here and broken down by the following categories:
+
+* [Cryptographic and Security](#cryptographic-and-security-functions)
+* [Date](#date-functions)
+* [Dictionaries](#dictionaries-and-dict-functions)
+* [Encoding](#encoding-functions)
+* [File Path](#file-path-functions)
+* [Kubernetes and Chart](#kubernetes-and-chart-functions)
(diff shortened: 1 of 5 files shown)


content/en/docs/chart_template_guide/function_list.md
```diff
@@ -0,0 +1,1951 @@
...
+### now
+
+The current date/time. Use this in conjunction with other date functions.
+
+### ago
+
+The `ago` function returns duration from time.Now in seconds resolution.
+
+```
+ago .CreatedAt"
```
technosophos, line 977: Extra quotation mark here

mattfarina:
Good catch on the `tpl` function. I found a few more I missed. I'll do that as a follow-up PR.

content/en/docs/chart_template_guide/function_list.md
mattfarina: Fixed. Thanks

content/en/docs/chart_template_guide/function_list.md
```diff
@@ -0,0 +1,1951 @@
...
+The `pluck` function makes it possible to give one key and multiple maps, and
+get a list of all of the matches:
+
+```
+pluck "name1" $myDict $myOtherDict
+```
+
+The above will return a `list` containing every found value (`[value1 otherValue1]`).
+
+If the give key is _not found_ in a map, that map will not have an item in the
```
glensc, line 1175: > If the give**n** key is _not found_ in a map

glensc:
The outcome is at this address: https://helm.sh/docs/chart_template_guide/function_list/

glensc:
@mattfarina I noticed that not all headings are listed in the toc beginning of the page:

- Kubernetes and Chart Functions
- File Functions

PR #701, opened 2020-06-16
[#553] add prev and next links to docs

### LINKED ISSUE #553: next button in every page
Hi,
Moving from one page in the docs to the next is very frustrating.
I'm missing two things:
  1. A link in the end of each page directing me to the next page.
  2. A sidebar that can lead to any page. The current sidebar only covers groups of topics and not specific ones.

flynnduism:
Adds auto-populated links to the end of each docs page:

Screen Shot 2020-06-16 at 10 01 12 AM

---

Closes #553.

### DIFF
--- a/themes/helm/assets/sass/docs-content.scss
+++ b/themes/helm/assets/sass/docs-content.scss
@@ -108,6 +108,18 @@
     }
   }
 
+  .prev-next {
+    border-top: 2px solid lighten($navyl, 33.33%);
+    padding: 1.5rem 0 0;
+    margin-top: 5rem;
+
+    a {
+      display: inline-block;
+      min-width: 66.67%;
+      padding: 1rem 0.5rem;
+    }
+  }
+
   blockquote {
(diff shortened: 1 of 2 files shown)


bridgetkromhout:
Hi, @flynnduism! This is awesome. I have a question - how does the auto-linking work? I picked a few arbitrary pages to check, and I'm baffled at the choices it's giving. For example, https://deploy-preview-701--helm-merge.netlify.app/docs/chart_best_practices/values/ - here's an example of what I thought would be linked, versus what is actually linked:

image

flynnduism:
Investigating this, I cannot see any logical sort order or sequence applied by Hugo here. Will update this PR to enforce some kind of better sorting.

lucperkins:
Personally, I'm not aware of the mechanism in use here. Have you tried using just plain `.Next` and `.Prev`?

flynnduism:
Updating this with `. PrevInSection` and `.NextInSection` resolved the page sequence issues. Thanks

PR #714, opened 2020-06-19
Replace  zh. toml  with the latest en.toml

arunfung:
Replace  zh. toml  with the latest en.toml

Signed-off-by: arunfung <arunfung@gmail.com>


arunfung:
@flynnduism @friederbluemle

Issue #728, opened 2020-06-24
Document $ in flow control


mattfarina:
When the `with` or `range` functions are used `.` changes its value. If one needs to access the root data object they need to use `$` instead. This should be documented in the flow control docs at https://helm.sh/docs/chart_template_guide/control_structures/

pankajme7:
@mattfarina Yes, use of .  changes the value. Should I add this information under "Modifying scope using with" subsection.

mattfarina:
@pankajme7 sure. Please also talk about it in loops where the same thing happens.

pankajme7:
@mattfarina 
I have done the changes as mentioned above. I have pushed it to a branch in my forked project. https://github.com/pankajme7/helm-www/commit/5342cc3b55daa223e03f4444a5dcbb4ff97240ad

Can you please guide me on how I can create PR for the same and link it with this issue.

mattfarina:
@pankajme7 The GitHub docs do have a good guide on turning a branch into a PR [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request). Just make sure your commits have git signoff (this is different from crypto signing) using the --signoff flag. This is for the DCO.

pankajme7:
@mattfarina 
Thanks for the help. I have created a PR for the same.

bridgetkromhout:
Thanks, @pankajme7! We'll wait for @mattfarina to confirm that this issue is resolved.

pankajme7:
Welcome, @bridgetkromhout

mattfarina:
The PR satisfied what I was looking for.

PR #737, opened 2020-07-01
Added documentation regarding use of $ for accessing objects from parent scope.

pankajme7:
Updated control flow documents to include information on how to access parent object using $. Inside 'with' and 'range'.

Signed-off-by: Pankaj Mehra <ppankajmehra050@gmail.com>


pankajme7:
Fix for issue #728

Issue #738, opened 2020-07-01
standardize on console instead of shell


bridgetkromhout:
As discussed in https://github.com/helm/helm-www/pull/732#discussion_r448437222, we propose standardizing on "shell" instead of "console" throughout the docs. This is a minor change for consistency and should not materially affect display.

bridgetkromhout:
After looking more at the docs, I begin to think it would be clearer and less confusing for the reader if they see a `$` everywhere they're expected to type something. This means we'd want to use "console".

Example: 
```console
$ helm repo update              # Make sure we get the latest list of charts
$ helm install stable/mysql --generate-name
Released smiling-penguin
```

Using "shell" gives less-formatted output on github:

```shell
$ helm repo update              # Make sure we get the latest list of charts
$ helm install stable/mysql --generate-name
Released smiling-penguin
```

GitHub renders this differently than we do on the website: https://github.com/helm/helm-www/blob/master/content/en/docs/intro/quickstart.md#install-an-example-chart vs https://helm.sh/docs/intro/quickstart/#install-an-example-chart

Another issue is that if we use "shell" with markdownlint, then all the `$` entries will be flagged as an error. As there is no display difference between the two on the website, but we risk making the docs more confusing with the "shell" option and also lose some formatting on GithHub, I'm now thinking we should instead use "console" in all cases (and add the `$` where missing). Input welcome.

bridgetkromhout:
When @mattfarina is available it would be good to get his input here.

hickeyma:
My experience with markdown is console and `$`.  For example:

```console
$ helm list -n test
NAME           NAMESPACE REVISION  UPDATED                                 STATUS    CHART            APP VERSION
[....]
```

bacongobbler:
For some historical reference...

The triple backtick convention was not a Markdown standard at first. Github expanded upon the Markdown spec and introduced the triple backtick for github comments to embed console output in a comment. This was commonly referred to as "Github-flavoured markdown".

Their convention was

```
```console
```

For shell sessions.

`shell` was introduced for shell scripts. If you're trying to fence a shell script, you should use `shell`.

Consider the syntax highlighting output between the two shell sessions here:

shell:

```shell
$ helm list --namespace=test    # a comment
NAME           NAMESPACE REVISION  UPDATED                                 STATUS    CHART            APP VERSION
[....]
```

console:

```console
$ helm list --namespace=test    # a comment
[...]

PR #744, opened 2020-07-06
Standardize on console vs shell

bridgetkromhout:
Fixes https://github.com/helm/helm-www/issues/738

While at first I thought we'd standardize on "shell", there are compelling reasons to go with "console" (which I've written up in https://github.com/helm/helm-www/issues/738#issuecomment-654447912). 

Let's wait to merge until @mattfarina can weigh in, as he had an opinion on this. 

### DIFF
--- a/content/en/docs/community/release_checklist.md
+++ b/content/en/docs/community/release_checklist.md
@@ -30,27 +30,27 @@ named, use a command like `git remote -v` to find out.
 If you don't have an upstream remote, you can add one easily using something
 like:
 
-```shell
-git remote add upstream git@github.com:helm/helm.git
+```console
+$ git remote add upstream git@github.com:helm/helm.git
 ```
 
 In this doc, we are going to reference a few environment variables as well,
 which you may want to set for convenience. For major/minor releases, use the
 following:
 
-```shell
-export RELEASE_NAME=vX.Y.0
-export RELEASE_BRANCH_NAME="release-X.Y"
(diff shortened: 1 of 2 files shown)


mattfarina:
I wanted to put this somewhere in writing. There are two places the code is displayed. First and primarily, there are the helm.sh docs. Secondarily there is GitHub. These use two different systems for highlighting.

Our Hugo setup uses https://github.com/yuin/goldmark-highlighting. Following the rabbit hole I see that `shell` is an alias for `bash`. Console is an alias for nothing. If you look at the [current release checklist in shell](https://helm.sh//docs/community/release_checklist/) you'll see highlighting. The [new one you won't](https://deploy-preview-744--helm-merge.netlify.app/docs/community/release_checklist/). If you look at the source, the new console highlighting is marked as fallback type.

GitHub does it differently and this is where I originally looked. Github uses https://github.com/github/linguist. There `console` is an alias for `bash session` otherwise known as files with a `.sh-session` extension. `shell` is for shell scripts.

I would suggest that the web docs, instead of GH, is the primary source. To make things best viewed there, shouldn't we use `shell`?

bridgetkromhout:
@mattfarina is correct; I was looking at the differences on GitHub, but as he points out, the differences on Hugo are far more important. My changes in this PR would move us (in the website display) from code which looks good:
image

to code which isn't formatted as nicely:
image


I agree that this is a non-starter. Closing.

PR #784, opened 2020-08-27
Clarify how define produces output

bridgetkromhout:
I wanted to capture the clarification in https://github.com/helm/helm/issues/7961#issuecomment-620143810 in the docs. Input from @technosophos and others welcome as to whether there is a better way or place to explain this.


technosophos (approved):
Good clarification.

PR #790, opened 2020-08-29
Replace the path with XDG_ prefix with helm env variables in docs

### LINKED ISSUE #336: References to XDG paths is insufficient
In multiple places in the documentation there are references to XDG paths.
For example it says in https://v3.helm.sh/docs/topics/charts/#chart-starter-packs:
 
"Currently the only way to add a chart to $XDG_DATA_HOME/helm/starters is to manually copy it there. In your chart’s documentation, you may want to explain that process."

So I can imagine a user would do something like:
```
cp ./myStarter $XDG_DATA_HOME/helm/starters
```
The problem is that XDG variables don't need to be set (when not set they use a default path location). So when trying to write or read from the path specified in the docs, things often fail because the XDG variable is not set.  In the example above, the copy command would become:
```
cp ./myStarter /helm/starters
```
which will normally fail with a permission error.

And once the user figures out what is going on, they somehow have to know what when these XDG vars are not set they default to some location that is OS dependant to boot.  Personally, I use `helm -h` to look at what the default XDG locations are.

The docs needs to provide some guidance when pointing the user to XDG paths.

rudeigerc:
Signed-off-by: Yuchen Cheng <rudeigerc@gmail.com>

Fixes #336 

According to helm/helm#8638, I replace the path with `XDG_` prefix with `helm env` variables in the docs.

I discovered that the cached index files were installed in `$(helm env HELM_CACHE_HOME)/repository` instead of `$XDG_CACHE_HOME/helm/repository/cache/` mentioned in the original docs, so I'm not quite sure which path is correct.

### DIFF
--- a/content/en/docs/topics/chart_repository.md
+++ b/content/en/docs/topics/chart_repository.md
@@ -308,5 +308,5 @@ latest chart information.
 
 *Under the hood, the `helm repo add` and `helm repo update` commands are
 fetching the index.yaml file and storing them in the
-`$XDG_CACHE_HOME/helm/repository/cache/` directory. This is where the `helm
+`$(helm env HELM_CACHE_HOME)/repository` directory. This is where the `helm
 search` function finds information about charts.*
--- a/content/en/docs/topics/charts.md
+++ b/content/en/docs/topics/charts.md
@@ -1108,7 +1108,7 @@ The `helm create` command takes an optional `--starter` option that lets you
 specify a "starter chart".
 
 Starters are just regular charts, but are located in
-`$XDG_DATA_HOME/helm/starters`. As a chart developer, you may author charts that
+`$(helm env HELM_DATA_HOME)/starters`. As a chart developer, you may author charts that
 are specifically designed to be used as starters. Such charts should be designed
 with the following considerations in mind:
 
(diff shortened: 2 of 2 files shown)


bridgetkromhout:
>  I'm not quite sure which path is correct.

@marckhouzam Can you review this PR? Thanks.

technosophos:
I feel uncomfortable adding UNIX-specific commands into the documentation in this way. Why not just change it to $HELM_CACHE_HOME and explain once how a user can get that information.

content/en/docs/topics/chart_repository.md
```diff
@@ -308,5 +308,5 @@ latest chart information.
 
 *Under the hood, the `helm repo add` and `helm repo update` commands are
 fetching the index.yaml file and storing them in the
-`$XDG_CACHE_HOME/helm/repository/cache/` directory. This is where the `helm
+`$(helm env HELM_CACHE_HOME)/repository` directory. This is where the `helm
```
mattfarina, line 311: Helm is used in non-POSIX systems like Windows. Commands that generally target that Linux like environments are something we try to avoid. In #780 I changed the variable name to the Helm one. This isn't the best solution those environment variables may be empty and using the defaults.

Can we get a solution that works for both Windows and Linux/Mac? It could possible tell people who to get the value for the environment variables and then show where something is.

mattfarina (changes requested):

PR #796, opened 2020-09-01
Add Chinese docs for faq and glossay

Zhanxueyou:

### DIFF
--- a/content/zh/docs/faq/_index.md
+++ b/content/zh/docs/faq/_index.md
@@ -1,0 +1,408 @@
+---
+title: "常见问答(FAQ)"
+weight: 8
+---
+
+# 常见问答
+
+> Helm 2和Helm 3的主要区别是什么?  
+> 该页面为最常见的问题提供帮助。
+
+**我们喜欢您的帮助** 使得该文档变得更好。添加、修改或移除信息,
+ [提问题](https://github.com/helm/helm-www/issues)或者向我们提交pull request。
+
+## 从Helm 2发生的变化
+
+这里会展示一个对Helm3主要引入的变化的详尽列表。
+
+### 移除了Tiller
+
+&emsp;&emsp;在Helm 2的开发周期中，我们引入了Tiller。Tiller在团队协作中共享集群时扮演了重要角色。
+它使得不同的操作员与相同的版本进行交互称为了可能。
+
+&emsp;&emsp;Kubernetes 1.6默认使用了基于角色的访问控制（RBAC），在生产环境对Tiller的锁定使用变得难于管理。
+由于大量可能的安全策略，我们的立场是提供一个自由的默认配置。这样可以允许新手用户可以乐于尝试Helm
+和Kubernetes而不需要深挖安全控制。 不幸的是这种自由的配置会授予用户他们不该有的权限。DevOps和SRE
+在安装多用户集群时不得不去学习额外的操作步骤。
+
+&emsp;&emsp;在听取了社区成员在特定场景使用Helm之后，我们发现Tiller的版本管理系统不需要依赖于集群内部用户去维护
+状态或者作为一个Helm版本信息的中心hub。取而代之的是，我们可以简单地从Kubernetes API server获取信息，
+在Chart客户端处理并在Kubernetes中存储安装记录。
+
+Tiller的首要目标可以在没有Tiller的情况下实现，因此针对于 Helm 3 我们做的首要决定之一就是完全移除Tiller。
+
+&emsp;&emsp;随着Tiller的消失，Helm的安全模块从根本上被简化。Helm 3 现在支持所有Kubernetes流行的安全、
+身份和授权特性。Helm的权限通过你的
+[kubeconfig文件](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/)进行评估。
+集群管理员可以限制用户权限，只要他们觉得合适，
+无论什么粒度都可以做到。版本发布记录和Helm的剩余保留功能仍然会被记录在集群中。
+
+### 改进升级策略: 三路策略合并补丁
+
+Helm 2 使用了一种双路策略合并补丁。在升级过程中，会对比最近一次的chart manifest和提出的
+chart manifest(通过`helm upgrade`提供)。升级会对比两个chart的不同来决定哪些更改会应用到Kubernetes资源中。
+如果更改是集群外带的（比如通过`kubectl edit`），则不会被考虑。结果就是资源不会回滚到之前的状态：
+因为Helm只考虑最后一次应用的chart manifest作为它的当前状态，如果chart状态没有更改，则资源的活动状态不会更改。
+
+现在Helm 3中，我们使用一种三路策略来合并补丁。Helm在生成一个补丁时会考虑之前老的manifest的活动状态。
+
+#### 示例
+
+让我们通过一些常见的例子来看看变化带来的影响。
+
+##### 回滚已经改变的活动状态
+
+你的团队正好在Kubernetes上使用Helm部署了生产环境应用。chart包含了一个部署对象使用了三套副本：
+
+```console
+$ helm install myapp ./myapp
+```
+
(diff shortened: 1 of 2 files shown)


content/zh/docs/faq/_index.md
```diff
@@ -0,0 +1,408 @@
...
+(也就是说运行`kubectl config view --minify`时会显示命名空间). 也就意味着您在执行`helm list`时必须提供
+ `--all-namespaces` 标识才能获得和Helm 2同样的结果。
+
+### 作为默认存储器的密钥
+
+In Helm 3, 密钥被作为[默认存储驱动](https://helm.sh/docs/topics/advanced/#storage-backends)使用。
+Helm 2默认使用ConfigMaps记录版本信息。在Helm 2.7.0中，新的存储后台使用密钥来存储版本信息，
+现在是Helm 3的默认设置。
+
+Helm 3默认允许更改密钥作为额外的安全措施在Kubernetes中和密钥加密一起保护chart。
```
bridgetkromhout, line 161: > Changing to Secrets as the Helm 3 default allows for additional security in protecting charts in conjunction with the release of Secret encryption in Kubernetes.

Does your translation say that Helm 3 uses Secrets as the default (instead of using the previous default which was ConfigMaps)?  Thanks.

content/zh/docs/faq/_index.md
```diff
@@ -0,0 +1,408 @@
...
+Helm 2中，如果没有提供名称， 会自动生成一个名称。在生产环境，这被证明是一个麻烦事而不是一个有用的特性。
+而在Helm 3中，如果 `helm install` 没有提供name，会抛异常。
+
+如果仍然需要一个自动生成的名称，您可以使用 `--generate-name` 创建。
+
+### 推送Chart到OCI注册中心
+
+这是一个Helm 3 中的实验性特性。使用时需要设置环境变量 `HELM_EXPERIMENTAL_OCI=1`。
+
+Chart仓库在较高层次上是一个存储和分发Chart的地址。Heml客户端打包并将Chart推送到Chart仓库中。
```
bridgetkromhout, line 235: > Heml

This should be `Helm`.

content/zh/docs/faq/_index.md
```diff
@@ -0,0 +1,408 @@
...
+
+Helm 插件仍然使用 `$HELM_HOME` 作为 `$XDG_DATA_HOME` 的别名，以便希望将`$HELM_HOME`作为过渡环境的变量来保证向后兼容性。
+
+一些新的环境变量也通过插件环境变量来兼容以下更改：
+
+- `$HELM_PATH_CACHE` 针对缓存路径
+- `$HELM_PATH_CONFIG` 针对配置路径
+- `$HELM_PATH_DATA` 针对data路径
+
+Helm插件期望支持Heml 3 就建议使用新的环境变量。
```
bridgetkromhout, line 313: "Heml" should be "Helm".

For this line, the English is: 
> Helm plugins looking to support Helm 3 should consider using these new environment variables instead.

It means "If a Helm plugin is going to support Helm 3, the plugin author should use these new environment variables instead of the old ones".

Is this what the translation says? Thanks.

content/zh/docs/faq/_index.md
```diff
@@ -0,0 +1,408 @@
...
+另外，其他一些重命名的命令提供了以下约定：
+
+- `helm inspect` -> `helm show`
+- `helm fetch` -> `helm pull`
+
+这些命令都保留了老的动词作为别名，因此您能够使用任意一种格式。
+
+### 自动创建namespace
+
+当创建了一个命名空间中不存在的版本时，Heml 2会创建一个命名空间。
```
bridgetkromhout, line 332: As mentioned elsewhere, Helm is the spelling we want in all cases, never Heml.

This line says that if the user tries to create a release in a namespace and the namespace does not already exist, Helm 2 will create the namespace automatically. Does the translation say that? Thanks.

content/zh/docs/faq/_index.md
```diff
@@ -0,0 +1,408 @@
...
+
+### 从Helm 2迁移后， `helm list` 仅显示部分（或者不显示）我的发布版本
+
+您很可能忽略了一个事实：Helm 3 现在使用集群的命名空间来确定版本范围。 这意味着所有涉及版本的命令您都必须：
+
+* 在活动的kubernetes上下文中需要依赖当前的命名空间 (如 `kubectl config view --minify` 命令所述)，
+* 使用 `--namespace`/`-n` 参数指定正确命名空间，或者
+* 对于 `helm list` 命令，指定 `--all-namespaces`/`-A` 参数
+
+这适用于 `helm ls`、 `helm uninstall` 以及其他所有涉及版本的 `helm` 命令。
```
bridgetkromhout, line 408: This line says:
>This applies to helm ls, helm uninstall, and all other helm commands referencing a release.

The "referencing a release" part means that this statement applies to all commands that are talking about a specific release. Does the translation explain this? Thanks.

content/zh/docs/glossary/_index.md
```diff
@@ -0,0 +1,128 @@
...
+## Chart依赖 (子chart)
+
+Chart可以依赖于其他的chart。 依赖可能会以以下两种方式出现:
+
+- 软依赖： 如果另一个chart没有在集群中安装，chart可能会无法使用。Helm未对这个案例提供工具。
+这个案例中，依赖会被分别管理。
+- 硬依赖： 一个chart可以包含 (在它的`charts/`目录中) 另一个它所依赖的chart。这个案例中，
+安装chart的同时会安装所有依赖。chart和它的依赖会作为一个集合进行管理。
+
+当一个chart(通过`helm package`)打包时所有的依赖都会和它绑定。
```
bridgetkromhout, line 30: Does this clarify that only "hard" dependencies and not "soft" ones will be packaged with the chart?

content/zh/docs/glossary/_index.md
```diff
@@ -0,0 +1,128 @@
...
+| upgrade    | release 2                                            |
+| upgrade    | release 3                                            |
+| rollback 1 | release 4 (但使用release 1的配置) |
+
+上面的列表阐明了发布版本号是怎样通过安装，升级和回滚递增的。
+
+## Helm库 (或SDK)
+
+Helm库（或SDK）涉及到go代码，可以直接与Kubernetes API服务交互进行安装、升级、查询
+以及移除Kubernetes资源。可以被导入到项目中来替换CLI使用的Helm库。
```
bridgetkromhout, line 108: > It can be imported into the project to replace the Helm library used by CLI.

The auto-translation here is wrong in English. Can you check to make sure this is correct? Here is what is being translated: "It can be imported into a project to use Helm as a client library instead of a CLI."

This means that Helm can be used as a client library instead of being used as a command. We're not replacing a specific library used by the CLI.

bridgetkromhout:
Hi, @Zhanxueyou! Thank you very much for this translation! We would like to merge this. I asked some questions - please check on those answers? Then we can merge. Thanks!

content/zh/docs/faq/_index.md
Zhanxueyou: Yes，I confirmed.

content/zh/docs/faq/_index.md
Zhanxueyou: OK, I got it. That's it.

content/zh/docs/faq/_index.md
Zhanxueyou: OK，sorry for that，I'll check it carefully in the future.

content/zh/docs/faq/_index.md
Zhanxueyou: Yes.

content/zh/docs/glossary/_index.md
Zhanxueyou: Yes, it's different from soft dependency.

content/zh/docs/glossary/_index.md
Zhanxueyou: I understood and checked again.

bridgetkromhout (approved):
Hi, @Zhanxueyou! Thank you very much for this translation.

PR #812, opened 2020-09-11
The Fedora PackageDB project has been retired in favor of the pkgdb2 …

Jmainguy:
…project


Jmainguy:
The current link to https://admin.fedoraproject.org/pkgdb/ 404's, this will change the link to go to a site that isnt 404. This is because the pkgdb has been deprecated. https://github.com/fedora-infra/packagedb/blob/develop/README

bacongobbler (approved):

PR #814, opened 2020-09-13
Use main branch for chartmuseum go mod

jdolitsky:
ChartMuseum has switched to use 'main' as the default branch name (vs. 'master'). This change ensures that Go projects import ChartMuseum Go libraries properly.


hickeyma (approved):
LGTM, thanks @jdolitsky

PR #818, opened 2020-09-15
Auto-generated docs update

bridgetkromhout:
I noticed that several recent docs updates were missing, so I thought it was perhaps time to refresh the auto-generated docs.

I'm not positive that the auto-generation switching to .md files is going to work in the site build, but it does work locally, so I'll submit it and see what the preview does.


bridgetkromhout:
The Netlify preview looks correct to me. I think merging this is probably a good thing since currently the website docs are lagging behind the CLI docs by some months.

flynnduism (approved):
Thank you for updating! LGTM

PR #868, opened 2020-09-25
Bump version for 3.3.4

marckhouzam:
While going through the release check-list I noticed the version in `config.toml` wasn't up to date.

I don't know what impact that has, but here is the fix.


marckhouzam:
Ok, I figured out that the version is used in the drop-down at the top of the docs page 😄

bacongobbler (approved):

PR #869, opened 2020-09-25
Add next release date as per HIP 0002

marckhouzam:
The following is a suggestion of how we could publish the next release's date.  But I don't know the site that well, so feel free to suggest a more appropriate location.

A new "Next Feature Release" section is added on the main page under the "Community" section.
The Release Checklist is also updated.
Japanese, Chinese and Korean will be shown in English until translated.

The next release is marked to be `3.5.0` for `January 13th, 2021` as suggest in the HIP review.
This PR, if approved, should only be merged once `helm 3.4.0` has been released.

This is what the new section would look like (notice the top left box):
NextReleaseHelm

### DIFF
--- a/themes/helm/layouts/index.html
+++ b/themes/helm/layouts/index.html
@@ -4,6 +4,7 @@ Helm
 
 {{ define "main" }}
 {{ $home := site.BaseURL }}
+{{ $nextversion := site.Params.nextversion }}
 
 <div id="helm" class="page-wrapper page-home">
   <header>
@@ -152,6 +153,12 @@ Helm
     <div class="columns is-desktop community-boxes">
 
       <div class="column is-one-third-desktop">
+        <section class="box">
+          <h1 class="title">{{ T "community_next_release" }}</h1>
+          <dt>{{ T "community_next_release_version" | markdownify }} {{ $nextversion.version }}</dt>
+          <dt>{{ T "community_next_release_date" | markdownify }} {{ $nextversion.date }}</dt>
+        </section>
+
         <section class="box">
           <h1 class="title">{{ T "community_events_title" }}</h1>
           <dl>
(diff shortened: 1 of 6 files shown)


themes/helm/layouts/index.html
```diff
@@ -152,6 +153,12 @@ <h2 class="has-text-centered">{{ T "community_title" }}</h2>
     <div class="columns is-desktop community-boxes">
 
       <div class="column is-one-third-desktop">
+        <section class="box">
+          <h1 class="title">{{ T "community_next_release" }}</h1>
+          <dt>{{ T "community_next_release_version" | markdownify }} {{ $nextversion.version }}</dt>
+          <dt>{{ T "community_next_release_date" | markdownify }} {{ $nextversion.date }}</dt>
```
mattfarina, line 159: I think the HTML tags should be different. I noticed there was not enough space below the dt and the bottom of the section in the visual display which lead me to this.

I think they needs to be a [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) that wraps `dt` aand `dd` tags. Or, something else.

In the following image see the difference in gap at the bottom of the box...
Screen Shot 2020-09-28 at 10 34 01 AM

marckhouzam: You are correct, I was missing the `<dl>` tags.
Instead of a two element in the list, I opted for a single element with a break.  That put less space between the version and the date lines.

What do you think?

mattfarina (approved):

marckhouzam:
Just a reminder that this should not be merged until after 3.4 is released.

Also I will add a page describing the release date "contract".

marckhouzam:
I've added a `release policy` page under the `Topics` section which details how release dates will be published.

hickeyma (approved):
LGTM, thanks @marckhouzam

mattfarina:
Since v3.4 is out we can publish this.

PR #870, opened 2020-09-26
Update supported versions

bridgetkromhout:
_What is this PR for?_
When Helm starts supporting k8s 1.19 as noted in https://github.com/helm/helm/pull/8807, we should note that in the version skew doc: https://helm.sh/docs/topics/version_skew/

_When do we merge this?_
This should be merged after the release of 3.4.0.


mattfarina (approved):

mattfarina:
Merging as v3.4.x is now out the door.

PR #877, opened 2020-10-01
Clarifying random strings between templates

bridgetkromhout:
Clarification for https://github.com/helm/helm/issues/6456.

### DIFF
--- a/content/en/docs/howto/charts_tips_and_tricks.md
+++ b/content/en/docs/howto/charts_tips_and_tricks.md
@@ -224,8 +224,8 @@ spec:
 [...]
 ```
 
-Each invocation of the template will generate a unique random string. This
-means that if it's necessary to sync the random strings used by multiple
+Each invocation of the template function will generate a unique random string.
+This means that if it's necessary to sync the random strings used by multiple
 resources, all relevant resources will need to be in the same template file.
 
 Both of these methods allow your Deployment to leverage the built in update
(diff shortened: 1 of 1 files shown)


content/en/docs/howto/charts_tips_and_tricks.md
```diff
@@ -224,6 +224,10 @@ spec:
 [...]
 ```
 
+Each invocation of the template will generate a unique random string. This
```
mattfarina, line 227: Where it says "template" I think you mean the "template function" `randAlphaNum`. It might be worth while to call it a template function.

bridgetkromhout: Thanks, @mattfarina - please see the update and let me know what you think.

scottrigby: Maybe we should clarify users would need to set a variable once to reuse the same random string even in the same file. Something like:

```yaml
{{ $random := randAlphaNum 5 | quote }}
---
kind: Deployment
spec:
  template:
    metadata:
      annotations:
        example: {{ $random }}
---
kind: Service
metadata:
  labels:
    example: {{ $random }}
```

mattfarina: It's also possible to generate a random value once and then reuse it across numerous templates. Should we document that, too? Maybe as follow-ups

mattfarina (approved):
lgtm

content/en/docs/howto/charts_tips_and_tricks.md
a-dot: > It's also possible to generate a random value once and then reuse it across numerous templates. Should we document that, too? Maybe as follow-ups

How?

PR #880, opened 2020-10-07
[minor edit] removing unnecessary special character

jmnote:
[minor edit] removing unnecessary special character


bridgetkromhout (approved):
Thanks for removing "�".

Issue #606, opened 2020-04-28
helm ship bouncing on waves = 100% CPU


planetf1:
I noticed that if I leave my browser open on helm.sh, that window will spin up to nearly 100% CPU.

This is seen on 
 - Safari Tech Preview Release 105 (Safari 13.2, WebKit 15610.1.10)
 - Google Chrome  Version 83.0.4103.23 (Official Build) beta (64-bit)

This is on a macbook pro 15 2016 model (4x2.7 i7)

mattfarina:
I'll add some additional data to this...

I loaded the homepage in the latest stable safari. It's is using ~38% CPU (as retrieved by Activity Monitor). Below is a picture of the dev tools during initial rendering.

![Screen Shot 2020-05-05 at 3 25 41 PM](https://user-images.githubusercontent.com/62991/81107307-f443e780-8ee4-11ea-8135-07dd8d932cad.png)

The CPU does taper off and if I run the dev tools timeline long term is notes a continued 22% usage.

@planetf1 how long does the 100% level stick around for?

planetf1:
On safari Version 13.1.1 (15609.2.7.1) I get approx 45% of one cpu when I check now - I've had a slight (beta) update since the above where I saw 100%. It's a 2016 vintage high-end MacBook.  I would still regard it as undesirable though.

kevindong:
To add on, the entire website is just laggy in general which is kinda ridiculous considering that this is meant to be a static documentation website. 

Aesthetics are certainly worth something, but it shouldn't get in the way of functionality.

PR #889, opened 2020-10-10
Improve homepage performance

kevindong:
# Why is this change being proposed?
The homepage performance on my computer (detailed below) is not great. There are two distinct issues:

* when scrolling, the usage of `background-attachment: fixed;` makes CPU usage skyrocket
* the non-transition animations keep CPU usage high even when the window isn't being interacted with (this is essentially what #606 reports)

# Changes being proposed
* Revert to using the default behavior for `background-attachment`
* Stop the boat from bobbing on the homepage (both the header section boat and the navbar badge boat)

# Evidence of these changes' efficacy
My test/personal machine is a base model 2015 13" Macbook Pro. In other words, dated and bordering on obsolete hardware. Quantitative evidence is below. Qualitatively, the homepage went from being very stuttery to being smooth. 

## Disabling non-transition animations
Before disabling animations, just having Chrome open on the homepage results in these resource usage levels:

no-movement-before

After disabling animations via commit eae35a7:

no-movement-after

## Stopping background image from being fixed
From rapidly scrolling up and down on the homepage after the header section, this is the resource usage level:

movement-before

After disabling the fixed background image via commit 8bd837e (which includes the changes made in commit eae35a7): 

movement-after

## The need for these changes
On my work machine (a base model 2019 16" Macbook Pro), these changes don't make much of a difference when the machine is not under a meaningful load. Obviously when it is under a load (e.g. I'm actively running applications, writing code, and referencing documentation), lower resource usage would be very appreciated.

### DIFF
--- a/themes/helm/assets/sass/helm-boat.scss
+++ b/themes/helm/assets/sass/helm-boat.scss
@@ -1,27 +1,3 @@
-@keyframes boat-bob {
-  0% {
-    bottom: 2vw;
-  }
-  50% {
-    bottom: 3.5vw;
-  }
-  100% {
-    bottom: 2vw;
-  }
-}
-
-@keyframes boat-badge-bob {
-  0% {
-    top: 0.425vw;
-  }
-  50% {
-    top: 0.75vw;
-  }
-  100% {
-    top: 0.425vw;
-  }
-}
-
 // boat full width default
 .boat {
   width: 100vw;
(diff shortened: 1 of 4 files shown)


bacongobbler:
I'm partially against this change because it completely disables it for all machines. I would be okay with a change that allows the animation to stop if the device is resource-constrained or disable animations. I'm not okay with removing animations entirely as that is part of the website's charm.

PR #892, opened 2020-10-13
Update template best practices regarding chompin

consideRatio:
Whitespace chomping are systematically done to the left in the templates output from `helm create`, and right whitespace chomping is reserved for when the initial content of a template file or defined template﻿ will follow. This PR will reflect that practice.

This PR also fixes smaller details such as to add YAML syntax highlighting in code blocks.


hickeyma (approved):
LGTM, thanks for the improvement to the doc @consideRatio.

I would like to get some more eyes on this to review before merging.

content/en/docs/chart_best_practices/templates.md
```diff
@@ -31,58 +31,81 @@ For that reason, _all defined template names should be namespaced._
...
 template code.
 
-```
-{{ if $foo -}}
-  {{- with .Bar }}Hello{{ end -}}
-{{- end -}}
+```yaml
+{{- range .Values.elements }}
+  {{- print "my element is" . }}
+{{- end }}
```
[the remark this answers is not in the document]
mattfarina: I think the point is useful. The use of indentation.

The old example used both `if` and `with` which is a bad example, too. What about using something like...

```
{{- with .Values.ingress.annotations }}
annotations:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

This is right from generated templates done by `helm create`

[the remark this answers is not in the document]
consideRatio: I think the docs point where this example resides, was orginally to suggest this kind of practice was acceptable, which i personally don't find it to be:

```yaml
annotations:
{{- if something }}
  {{- if something-else }}
    {{- if something-else-again }}
  my.silly.org/annotation: "really silly"
    {{- end }}
  {{- end }}
{{- end }}
```

I think one should always do this instead.

```yaml
annotations:
  {{- if something }}
  {{- if something-else }}
  {{- if something-else-again }}
  my.silly.org/annotation: "really silly"
  {{- end }}
  {{- end }}
  {{- end }}
```

I don't know if there is clear consensus on this kind of practice in the Helm docs anywhere, except for the _helpers.tpl file is mostly following the latter.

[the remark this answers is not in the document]
consideRatio: My recommendation is that this entire example is removed.

> Blocks (such as control structures) may be indented to indicate flow of the
template code.
>
>```yaml
>{{ if $foo -}}
>   {{- with .Bar }}Hello{{ end -}}
>{{- end -}}
>```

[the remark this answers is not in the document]
consideRatio: Or, replaced with an example that one should NOT try to indent templates based on nested template logic, but stay on the indentation level based on the YAML.

DO THIS

```yaml
annotations:
  {{- if something }}
  {{- if something-else }}
  {{- if something-else-again }}
  my.silly.org/annotation: "really silly"
  {{- end }}
  {{- end }}
  {{- end }}
```

DO NOT DO THIS:
```yaml
annotations:
  {{- if something }}
    {{- if something-else }}
      {{- if something-else-again }}
  my.silly.org/annotation: "really silly"
      {{- end }}
    {{- end }}
  {{- end }}
```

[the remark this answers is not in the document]
consideRatio: I conclude that the most reasonable idea is to remove this section. A clear motivation for this is that it is currently having a comment after it saying

Blocks (such as control structures) may be indented to indicate flow of the
template code.

> However, since YAML is a whitespace-oriented language, it is often not possible for code indentation to follow that convention.

See dd2b5ff for the change resolving this discussion in my mind - it removes this section. In my mind, it is a suggestion of a best practice I don't agree with, and, which only is relevant in very rare situations.

consideRatio:
@mattfarina we discussed a section and I have now landed in a concrete idea: to remove the section which I question to be a best practice. This was done in dd2b5ff. I now consider this ready for merge in my mind.

bridgetkromhout:
Hi, @mattfarina @hickeyma - thoughts on this PR? Should we have the merge conflict corrected so it can be merged, or are we past wanting this update? Thanks.

consideRatio:
@bridgetkromhout thanks for the triage work! I rebased it in case its considered a positive change.

content/en/docs/chart_best_practices/templates.md
```diff
@@ -31,62 +31,73 @@ For that reason, _all defined template names should be namespaced._
 Correct:
 
 ```yaml
-{{- define "nginx.fullname" }}
-{{/* ... */}}
-{{ end -}}
+{{- define "nginx.fullname" -}}
+# ...
```
bacongobbler, line 35: It's not totally clear here why you swapped out the Go template comment syntax with YAML comments. The former will be removed from the rendered output, while the latter will not. It also introduces the user to the text/template syntax.

https://golang.org/pkg/text/template/#hdr-Actions

Can you please revert this as well as any other similar changes in this PR?

content/en/docs/chart_best_practices/templates.md
```diff
@@ -31,62 +31,73 @@ For that reason, _all defined template names should be namespaced._
...
-```
 
-Incorrect:
-```
-{{.foo}}
-{{print "foo"}}
-{{-print "bar"-}}
+```yaml
+{{- /* Conditionally create service account */}}
+{{- if .Values.serviceAccount.create }}
```
bacongobbler, line 61: This will error out if `serviceAccount` is a null value. I would refrain from using nested values in examples. Especially since the ONLY thing this example is trying to convey is "template formatting".

The original examples were just fine IMO. Why are we changing this?

content/en/docs/chart_best_practices/templates.md
```diff
@@ -31,62 +31,73 @@ For that reason, _all defined template names should be namespaced._
...
 
-Templates should chomp whitespace where possible:
+Incorrect:
 
 ```yaml
-foo:
-  {{- range .Values.items }}
-  {{ . }}
-  {{ end -}}
+{{-/* This comment is even a syntax error */}}
```
bacongobbler, line 70: It's probably best to not overload the user with more than one topic at one time. Please keep this focused on the content that is trying to be conveyed. To clarify, here we only want to point out that "Template directives should have whitespace after the opening braces and before the closing braces". Anything more than that can lead to confusion for the reader.

I would also remove the dashes from this example. Again we're only trying to introduce the bare minimum here to keep the content clear and concise. Whitespace chomp syntax is introduced further in the article.

content/en/docs/chart_best_practices/templates.md
```diff
@@ -31,62 +31,73 @@ For that reason, _all defined template names should be namespaced._
 Correct:
 
 ```yaml
-{{- define "nginx.fullname" }}
-{{/* ... */}}
-{{ end -}}
+{{- define "nginx.fullname" -}}
```
bacongobbler, line 34: IMO I would remove whitespace chomping syntax until after it's been introduced to the reader. Introducing it before the user knows what it's supposed to do will only lead to questions/confusion.

content/en/docs/chart_best_practices/templates.md
```diff
@@ -147,17 +158,23 @@ metadata:
...
+Helm template comments:
+
 ```yaml
+{{- /* This is a comment (with chomping to the right). */ -}}
+type: frobnitz
+
 {{- /*
-This is a comment.
+This is a multiline comment (with no chomping to the right).
 */}}
```
bacongobbler, line 176: why is there space between the comment syntax on line 174 and not here?

bacongobbler (changes requested):

content/en/docs/chart_best_practices/templates.md
consideRatio: It is required syntax, but not at all obvious and I bet more than I have run into this kind of syntax error. I added a commit warning about it.

>When using template comments, watch out for a common syntax error.
>
>Correct:
>
>```
>{{- /* Correct spacing with whitespace chomping*/ -}}
>{{/* Correct spacing without whitespace chomping */}}
>```
>
>Incorrect:
>
>```
>{{-/* Incorrect spacing with whitespace chomping */-}}
>{{ /* Incorrect spacing without whitespace chomping */ }}
>```

content/en/docs/chart_best_practices/templates.md
consideRatio: > It's probably best to not overload the user with more than one topic at one time.

I agree - example minimized to a single line.

content/en/docs/chart_best_practices/templates.md
consideRatio: We are trying to convey a best practice in this article, and leading by examples not following such practices only to correct them in the end wouldn't be very effective I believe - and could potentially be counter-effective.

As an example on why I believe that, consider the science on how to debunk a myth. It has been found that it is bad to first declare the myth. Instead one should first declare how it really is, then warn that a misconception is about to be mentioned, mention the myth/misconception, and finally explain why it was a misconception/myth.

Similarly, I don't think we should lead with examples breaking the recommended practice only to later suggest how the practice should be. I know comparing this to debunking myths isn't correct, and I know there is value to increase the complexity step by step. With that said based on my experience as a teacher, from what I've learned from research, and using this documentation as a Helm chart developer, I believe the biggest concern is to be inconsistent. 

So my suggestion would be to either lead with the whitespace chomping section so it can be introduced directly, or to be consistent in the whitespace chomping practices until the practices are introduced.

content/en/docs/chart_best_practices/templates.md
consideRatio: Okay I'll retain template comments instead of YAML comments. I made the switch because I felt it increased readability, partially because syntax highlighting and partially because it felt like it communicated better that it was placeholder content.

content/en/docs/chart_best_practices/templates.md
consideRatio: No more serviceAccount.create or nested values! Original example restored!

PR #902, opened 2020-10-20
Fixing paths

bridgetkromhout:
I had an error creep in that apparently I didn't correct in all cases. Noticed this because it was appearing in translations.


hickeyma (approved):
LGTM, thanks

mattfarina (approved):
lgtm

PR #920, opened 2020-10-29
[ja] Translate community/localization in Japanese

dragoneena12:
I translated `community/localization.md` to Japanese.


bridgetkromhout (approved):
Thank you, @dragoneena12!

PR #921, opened 2020-10-30
docs(helm): Generate docs for 3.4.0

marckhouzam:
Here are the docs for 3.4.0.
The PR is pretty big because helm no longer shows kubernetes global flags, so each command was affected.


bridgetkromhout (approved):
LGTM

PR #922, opened 2020-10-30
Mention to run proper version to generate docs

marckhouzam:
I thought it may be of value to mention that when generating the helm docs, the specific helm version should be used.

For instance, my version helm is often point to the latest master, which would not represent the exact commands that need to be generated for a specific version.


bridgetkromhout (approved):
Good point!

PR #925, opened 2020-11-03
Add Kubernetes API doc

hickeyma:
Follow on from Kubernetes API doc which was added in https://github.com/helm/helm/pull/8135/. This PR adds the hook to render the doc for the web docs.

PR #930, opened 2020-11-05
Adding deprecated APIs link to sidebar

bridgetkromhout:
Addition for https://github.com/helm/helm-www/pull/925

Co-authored-by: Martin Hickey <martin.hickey@ie.ibm.com>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -63,46 +63,52 @@ url = "https://v2-14-0.helm.sh/docs"
     weight = 12
     parent = "using_helm"
   [[menu.main]]
+    name = "Deprecated Kubernetes APIs"
+    url = "deprecated-kubernetes-apis"
+    identifier = "deprecated-kubernetes-apis"
+    weight = 13
+    parent = "using_helm"
+  [[menu.main]]
     name = "Kubernetes Distro Notes"
     url = "kubernetes-distribution-guide"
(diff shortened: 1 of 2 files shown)


bacongobbler (approved):

bridgetkromhout:
Okay, @hickeyma - see https://deploy-preview-930--v2-helm.netlify.app/docs/ for the preview - is this what you envisioned?

image

flynnduism (approved):

hickeyma (commented):
Thanks for pushing this @bridgetkromhout. Hmm, looking at it now, I am not so sure that it is the best place to put it. Do folks think it should be within "Using Helm" or on its own like "Developer Guide"? It is kind of a special/advanced topic.

bridgetkromhout:
>  I am not so sure that it is the best place to put it. Do folks think it should be within "Using Helm" or on its own like "Developer Guide"? It is kind of a special/advanced topic.

Anyone who uses Helm does need to pay attention to k8s API deprecation, because APIs going away will likely cause problems for an ever-increasing number of users. I think keeping it here seems okay until the next time we re-organize docs.

hickeyma:
I think that is reasonable @bridgetkromhout.

hickeyma (approved):
LGTM, thanks @bridgetkromhout

PR #940, opened 2020-11-07
[zh] Add Chinese docs

Zhanxueyou:


bridgetkromhout (approved):
Thank you, @Zhanxueyou! 

One note: we prefer https instead of http - we may change those links - but other than that, looks great!

PR #947, opened 2020-11-11
Adding release calendar redirect

mattfarina:
Helm adopted a release schedule and there is now a release
calendar. This provides a redirect to the calendar.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -39,3 +39,8 @@
   from = "/helm/v2/*"
   to = "/helm/v2"
   status = 200
+
+[[redirects]]
+  from = "/calendar/release"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -39,3 +39,8 @@
   from = "/helm/v2/*"
   to = "/helm/v2"
   status = 200
+
+[[redirects]]
+  from = "/calendar/release"
```
scottrigby, line 44: I wondered if the trailing slash mattered here, but this looks good per https://docs.netlify.com/routing/redirects/

scottrigby (approved):
I think this will be helpful for end users to know what to expect when. 💯

bridgetkromhout:
Checked https://deploy-preview-947--helm-merge.netlify.app/calendar/release and it works as advertised.

PR #956, opened 2020-11-16
Update README to use new --generate-headers

marckhouzam:
To generate helm command documentation, the new `--generate-headers` flag allows to avoid having to add the headers back by hand.  See https://github.com/helm/helm/pull/8961

This PR updates the doc that explains how to generate that documentation.

This new flag will be part of helm 3.5.  So we may want to wait until it is release to merge this.  However, I'm not sure the docs will need to be regenerated for helm 3.4.1 or 3.4.2, so it may be ok to merge this change now, I'm not sure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,7 @@ To update these docs, you'll need to:
 
 1. Delete all plugins currently installed with `helm plugin uninstall`
 2. Navigate to `content/en/docs/helm/`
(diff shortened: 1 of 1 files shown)


bacongobbler (approved):
The next time we will need to update the docs will be Helm 3.5, so it seems fine to me to merge. I'll hold off on merging to let others weight in here, but LGTM.

bridgetkromhout:
I agree we're unlikely to need to generate the docs again before 3.5. I also think that the only people likely to be following those directions are weighing in on this PR, so merging it is perfectly safe.

PR #964, opened 2020-11-30
Add link to OCI HIP from Registries page

kleintech:
👋🏻 Hi there! 

I personally found the [Registries ](https://helm.sh/docs/topics/registries/) page on my own and found it very helpful, but ended up shying a way a bit when I found out it was still an experimental feature. There wasn't much there that explained why that was and the OCI HIP was in the community repo. I think it prudent to add a link to the OCI HIP in helm/community to lead folks there for more info.


bacongobbler (approved):
this is a great addition to the docs. Thanks @kleintech!

PR #987, opened 2020-12-04
Update charts.md

indrojeet1061:
Removed documentation for "enabled" field in charts.md file. As I understand, this field was might have been exposed by mistake. Setting it to a boolean value does not enable for disable chart loading.


mattfarina (approved):
lgtm

PR #999, opened 2020-12-17
docs: Add pre-req for running tests

sbose78:
* Golangci-lint is needed for running all tests.

We can update the docs after https://github.com/helm/helm/pull/7695 is merged.

### DIFF
--- a/content/en/docs/community/developers.md
+++ b/content/en/docs/community/developers.md
@@ -29,11 +29,21 @@ If required, this will first install dependencies, rebuild the `vendor/` tree,
 and validate configuration. It will then compile `helm` and place it in
 `bin/helm`.
 
-To run all the tests (without running the tests for `vendor/`), run `make test`.
 
 To run Helm locally, you can run `bin/helm`.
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Thanks, @sbose78. Yes, we can update docs once that PR is merged, but it does not seem to be actively under development right now.

bridgetkromhout:
Hi, @sbose78 - it appears that PR did not get completed. If you would like to take it over and submit a workable version we can merge, then after the code is updated we can consider docs updates. But for now I'll close this. Thanks!

sbose78:
Sure, but as of now, explicit installation of golangci-lint is required, hence thought we should document it :-)

bacongobbler:
I think @bridgetkromhout was stating that we would happily accept PRs to document this. The problem is that there is no PR currently in the queue which is ready to be merged.

If you re-submit a new PR which can be merged, we'd happily accept it.

bacongobbler (approved):

PR #1000, opened 2020-12-22
docs: refer to Tanka and the helm-tanka plugin

Duologic:
Adds 2 references: 

* helm-tanka plugin to consume Tanka/Jsonnet with Helm
* Tanka, which can consume Helm Charts


jdolitsky (approved):

bridgetkromhout:
Thanks, @Duologic!

PR #1005, opened 2021-01-13
Add Move2Kube to additional tools

ashokponkumar:
[Move2Kube](https://konveyor.io/move2kube/) helps users create helm charts for existing projects, and even parameterizes common variables. It supports source platforms which are not yet containerized, supports cloud foundry, docker compose and more. It can also support Kubernetes artifacts as source.


bridgetkromhout (approved):
Thanks!

PR #1010, opened 2021-01-26
Document dependencies from OCI registry

bdabelow:
Document how to specify dependencies in Chart.yaml so that depencency update can pull them.

This resolves a confusion I had in https://github.com/helm/helm/issues/9267.


jdolitsky (approved):
👍

PR #1016, opened 2021-02-04
Adjust docs to reflect Artifact Hub & helm/charts updates

bridgetkromhout:
Adjustments to docs to reflect the following changes:

- Artifact Hub, not the helm/charts repo, is where people can find charts
- There is not an active "Kubernetes Charts" project
- client-side SSL is not a work in progress anymore

There is not a new location for "common helm helper chart" and it's listed as deprecated, so that section needs overhaul, but I made a minimal change to acknowledge that it's not currently the recommendation.

While many other improvements are warranted, this is fairly low-hanging fruit to get in so the docs are more correct.


technosophos (approved):
LGTM. Thanks for sorting through all of that.

Issue #965, opened 2020-12-01
Use a helm repo that is not full of deprecated charts as the example


craiglpeters:
It feels strange to point to https://charts.helm.sh/stable in https://helm.sh/docs/intro/quickstart/#initialize-a-helm-chart-repository as it is full of charts marked DEPRECATED. It is confusing to a first time user. Should probably just stick the boring/canonical ngnix-ingress chart example and add https://helm.nginx.com/stable

bridgetkromhout:
Thanks! Many places in the docs need to be updated (this all changed in mid-November).

bacongobbler:
fixed in #1017. When the helm/examples repo is ready, we can replace the current examples with that one. But #1017 removes the deprecation warnings for first-time users so this gets us hobbling along.

PR #1017, opened 2021-02-06
Update quickstart guide to use non-deprecated repo

armandgrillet:
This allows the quickstart guide to not show deprecation messages to the user when following it. The issue was raised in #965.

Using `bitnami` as repo is subject to edits, it seems like the closest to what https://charts.helm.sh/stable was and it is the first Helm-related repository containing charts when using this list: https://github.com/topics/charts?o=desc&s=stars

Translations remain to be done but I believe a first pass should be given to review the repo used.

### DIFF
--- a/content/en/docs/intro/quickstart.md
+++ b/content/en/docs/intro/quickstart.md
@@ -40,38 +40,44 @@ Hub](https://artifacthub.io/packages/search?kind=0) for available Helm chart
 repositories.
 
 ```console
-$ helm repo add stable https://charts.helm.sh/stable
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Thanks, @armandgrillet! We're going to point the documentation to a small collection of example charts: https://github.com/helm/examples though we haven't gotten started yet - and we'll be happy to get PRs over there!

bacongobbler (approved):
Until that examples repo is ready, this will suffice. Bitnami was the original contributor for the helm charts repository and we've referenced their material in the past, so we can trust them as a good source for examples.

Thanks!

PR #1022, opened 2021-02-21
Improve wording

CorralPeltzer:
The word "container" in this context might lead to the misunderstanding that ConfigMaps create containers/pods.


bridgetkromhout (approved):
Hi, @CorralPeltzer - thank you for this clarification! I agree; "container" means a specific thing in the minds of many people, and it's good if we can disambiguate.

PR #1031, opened 2021-02-27
Fix console output

kit494way:
Template output is wrong.
A different value is used from the release name passed as a command line parameter.


bridgetkromhout (approved):
Thanks for improving the docs, @kit494way!

PR #1038, opened 2021-03-11
Fix menu in Edge and Chrome on Windows

### LINKED ISSUE #914: Docs: Console error showing up in Chrome
When I try opening the page located at [https://helm.sh/docs/topics/](https://helm.sh/docs/topics/) using Chrome, I get the error as shown in the screenshot.

Additionally, the page entries do not seem to show up under each section in the left navigation bar.

It seems to work fine in Firefox. I've double checked that there's no interfering plugin - tried in incognito as well. Also tried flushing the site data & cache, and did a hard reset.

In Chrome 86.0.4240.111 on Windows 10 (console trace on RHS):
image

In Firefox 82.0.1 on Windows 10:
image

Best,
Samir.

tastypackets:
This fixes an issue that prevents the menu from expanding on Chrome and Edge, reading the issue discussion it sounds like there is a goal to move away from this menu code and I'd agree that is the best long term solution. Currently it's pretty bad rendering, however this bug makes the menu completely broken on Windows Chrome & Edge and I think should be fixed before waiting for rewrite.

Fixes #914

Here is a gif showing what is broken:
Broken on edge

Here are gifs showing it working in all FireFox, Edge, and Chrome on Windows:
Works on firefox
Works on edge
Works on chrome

### DIFF
--- a/themes/helm/assets/js/foundation.js
+++ b/themes/helm/assets/js/foundation.js
@@ -410,7 +410,7 @@
       // or make it an empty string:
       //    Foundation.global.namespace = '';
       //
-      //
+      Foundation.global.namespace = 'helm-www';
 
       // If the namespace has not been set (is undefined), try to read it out of the meta element.
       // Otherwise use the globally defined namespace, even if it's empty ('')


bacongobbler (changes requested):
Looks good, however it appears this removes an example within the doc comments. It’s probably best to just add this line after the comments so it reads better.

Can you please revert line 409 and add it to line 413? Thanks!

tastypackets:
> 
> 
> Looks good, however it appears this removes an example within the doc comments. It’s probably best to just add this line after the comments so it reads better.
> 
> Can you please revert line 409 and add it to line 413? Thanks!

👍

bacongobbler (approved):

PR #1041, opened 2021-03-13
Removed a word which is duplicated and added round bracket which is missed.

ByoungJoonIm:
Actually I don't exactly know which one is right about 'using'.


ByoungJoonIm:
Because I pused together, I'll recreate pull requests.

PR #1045, opened 2021-03-13
Change command to remove all files including tests

kit494way:
`rm -rf mysubchart/templates/*.*` keeps files in tests directory, and subsequent ` helm install --generate-name --dry-run --debug mychart/charts/mysubchart` fails.


hickeyma (approved):
Goof catch, thanks @kit494way

PR #1065, opened 2021-04-14
Updated Chinese translation

RockyFu267:
The translation "图片" is similar to the meaning of the picture, it will be more accurate to express with “镜像”.
@Zhanxueyou


bridgetkromhout (approved):
Thank you!

PR #1066, opened 2021-04-14
Spanish translation

jorgevillaverde:
Signed-off-by: jorgevillaverde <jorge.villaverde@toucansoft.io>

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -322,7 +322,7 @@ weight = 3
 
 [[languages.es.menus.main]]
 name = "Blog"
-url = "/blog"
+url = "https://helm.sh/blog"
 weight = 4
 
 [[languages.es.menus.main]]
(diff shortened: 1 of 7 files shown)


config.toml
```diff
@@ -295,4 +295,37 @@ weight = 4
...
-weight = 5
\ No newline at end of file
+weight = 5
+
+# Spanish
+[languages.es]
+title = "Helm"
+description = "Helm - El Administrador de Paquetes de Kubernetes."
+contentDir = "content/es"
+languageName = "Español"
```
bridgetkromhout, line 305: for consistency with the other languages, you could go with:

languageName = "Español Spanish"

(I agree this is not perfect and I suspect we may need to do some sorting soon.)

bridgetkromhout (approved):
Gracias!

PR #1069, opened 2021-04-14
feat(helm): optionally hide validation warnings

rcousineau-xandr:
accompanies https://github.com/helm/helm/pull/9604

PR #1071, opened 2021-04-15
Clarifying where CLI docs are edited.

bridgetkromhout:
Clarifying where CLI docs are edited, based on feedback in https://github.com/helm/helm-www/pull/1069#issuecomment-819933254.


bacongobbler (approved):
LGTM. Explicitly pointing out the exception to the rule is good here as a number of people have run into this issue. Hopefully this can prevent more cases in the future.

PR #1074, opened 2021-04-16
Language Menu Adjustments 

flynnduism:
This PR adjusts the CSS for the top menu bar as follows:

* restores the correct spacing following banner removal (#1054)
* makes use of the active class to highlight current language

In addition, this PR changes the way the languages menu is populated, sorting them in order of the language code (`en`, `ko`, `zh` etc) and removes the English language word for each language from the title.

---

Current menu screenshot:

Screen Shot 2021-04-16 at 1 16 27 PM

Proposed change screenshot:

Screen Shot 2021-04-16 at 1 15 21 PM

Screen Shot 2021-04-16 at 1 15 04 PM

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -122,8 +122,8 @@ weight = 5
 title = "헬름"
 description = "헬름 - 쿠버네티스 패키지 매니저"
 contentDir = "content/ko"
-languageName = "한국어 Korean"
-weight = 1
+languageName = "한국어"
+weight = 5
 
 [[languages.ko.menus.main]]
 name = "홈"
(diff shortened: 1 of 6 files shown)


flynnduism:
Note: the sorting technique used to generate this order is adding `weight` values in the [site config](https://github.com/helm/helm-www/pull/1074/files#diff-28043ff911f28a5cb5742f7638363546311225a63eabc365af5356c70d4deb77). I didn't want to do it this way, as it should just be automated to list the `en` codes alphabetically. 

```
[languages.zh]
title = "Helm"
description = "Helm - Kubernetes 包管理器"
contentDir = "content/zh"
languageName = "中文"
weight = 7
```

However after applying different sorting techniques to the Hugo templates to get the desired A-Z order, nothing worked as expected. Every different method I tried to sort with resulted in odd, random sequences 🤷‍♂️

e.g.`{{ range sort $siteLanguages "Lang" }}` would result in:

Screen Shot 2021-04-16 at 12 25 21 PM

bridgetkromhout (approved):
Thanks for your work on this! I think this looks great. Thanks.

PR #1082, opened 2021-04-25
fix Chinese translation

RockyFu267:
The meaning of these two sentences is reversed；

覆盖父chart中的值---->Parent Chart is overwritten.

顶层的值已经被子chart的值覆盖了---->
The Parent Chart is covered by the SubChart.


bridgetkromhout (approved):
Thank you for fixing this!

Issue #1079, opened 2021-04-23
New helm plugin - official doc


HamzaZo:
Hello!
I create a new helm v3 plugin to help the community to fetch values from an older release.  [helm-val](https://github.com/HamzaZo/helm-val)

I would like to know what are the criteria to add a plugin to the official [documentation](https://helm.sh/docs/community/related/#helm-plugins)?

Thanks

bacongobbler:
transferred to helm-www.

HamzaZo:
Thanks @bacongobbler for your answer

bridgetkromhout:
Thanks, @HamzaZo!

PR #1083, opened 2021-04-26
add new plugin

HamzaZo:
This PR adds a new plugin under plugins and other third party projects. Related to #1079


bridgetkromhout (approved):
Thanks, @HamzaZo!

PR #1084, opened 2021-04-27
[es] Translate docs/topics/library_charts to Spanish

jorgevillaverde:
- Translate docs/topics/library_charts.

Signed-off-by: jorgevillaverde <jorge.villaverde@toucansoft.io>


bridgetkromhout (approved):
Thanks!

PR #1094, opened 2021-05-07
Add example of --template usage

yariksheptykin:
Current documentation list all available template vars, but does not provide an example of a template string that demonstrates template substitution. I had to look it out in the unit test.


bacongobbler:
Hi @yariksheptykin. These docs are autogenerated from the Helm CLI help text. Any changes being made here will have to be made to the Help string first.

See https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/cmd/helm/version.go#L44-L50

Once your changes are merged into helm/helm and have been released, we can re-generate the documentation following https://github.com/helm/helm-www#updating-the-helm-cli-reference-docs

Hope this helps. Thanks!

PR #1095, opened 2021-05-07
Add Greek Language

nikfot:
This is the initial docs for greek language. It consists of the content directory for greek and the corresponding _index.md file, the gr.toml inside the i18n dir and the greek language added in the config.toml

Signed-off-by: Nikos Fotiou <nik_fot@hotmail.gr>


bridgetkromhout:
Hi, @nikfot! Thank you! This is very exciting! Two things:

1) Do you want to translate "Docs", "Charts", "Blog" so they show up in Greek on the front page? You can see the preview at https://deploy-preview-1095--helm-merge.netlify.app/gr/

2) Do not worry too much about the order in the menu - as explained in https://github.com/helm/helm-www/pull/1074#issuecomment-821580748 it needs manual adjustment in the weighting, which we have not properly documented. We can fix that later if you don't want to experiment with weighting to get it into the right sort order.

nikfot:
Hi, @bridgetkromhout ! I am very excited about joining the community as well! This is my first step to further contribution.

While there are greek words for "Docs", "Charts" and "Blogs", the fact is that in everyday work we use the term "Chart" in our team as well and the word "Blog" is used in greek sites to describe the blog. Maybe I could change the word "Docs" into a more greek word, to give a more greek-feel to the menu.

What I am talking about is that in some cases I chose everyday language used by colleagues over direct translation - which I though would sound funny if one of my colleagues came across it-. I will change the "Docs" word, and please let me know if you believe that direct translation would be more suitable.

bridgetkromhout:
> I chose everyday language used by colleagues over direct translation

That is wonderful! See, this is why translations by language speakers who also use the software are the best! 

Thanks so much for getting this started! I'm excited to see more. :)

nikfot:
> > I chose everyday language used by colleagues over direct translation
> 
> That is wonderful! See, this is why translations by language speakers who also use the software are the best!
> 
> Thanks so much for getting this started! I'm excited to see more. :)

I will get back with more soon! 
Thanks for pointing out important details to do this the right way.

PR #1103, opened 2021-05-15
Greek: Add translated files to docs

nikfot:
Added _index.md , conventions.md , dependencies.md in docs. 
_In the case of depenencies.md translation was a bit tricky so the term "chains" was used which is more accurate than the direct translation of "dependencies" in greek._


content/gr/docs/chart_best_practices/_index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: "Βέλτιστες Πρακτικές"
+sidebar: true
+weight: 4
+aliases: ["/docs/topics/chart_best_practices/"]
```
bridgetkromhout, line 5: The aliases are only used to redirect pages that have moved. In this case, this would accidentally redirect people looking for an English doc to this Greek page, so we definitely need to remove this line.

content/gr/docs/chart_best_practices/dependencies.md
```diff
@@ -0,0 +1,53 @@
+---
+title: "Αλυσίδες Charts"
+description: "Καλύπτει τις βέλτιστες πρακτικές για τις αλυσίδες chart εντός ενός Chart."
+weight: 4
+aliases: ["/docs/topics/chart_best_practices/dependencies/"]
```
bridgetkromhout, line 5: Same as above - please remove this line.

content/gr/docs/chart_best_practices/conventions.md
```diff
@@ -0,0 +1,45 @@
+---
+title: "Γενικές Συμβάσεις"
+description: "Γενικές ΣΥμβάσεις για τα charts."
+weight: 1
+aliases: ["/docs/topics/chart_best_practices/conventions/"]
```
bridgetkromhout, line 5: Same as above - please remove this line.

content/gr/docs/chart_best_practices/_index.md
nikfot: Thanks for pointing that out @bridgetkromhout ! I fixed it.

bridgetkromhout (approved):
Thanks!

PR #1115, opened 2021-05-21
More grammar/style fixes in function_list

eric-unc:


content/en/docs/chart_template_guide/function_list.md
```diff
@@ -1891,15 +1889,12 @@ The basic comparisons are:
 - `>=`: greater than or equal to
 - `<=`: less than or equal to
 
-_Note, according to the Semantic Version specification pre-releases may not be
-API compliant with their release counterpart. It says,_
```
bridgetkromhout, line 1895: Quick question - why remove this line? Thanks.

eric-unc: Well there was no quote after, but in the next section it repeats this same line (minus "Note,") but actually gives the quote from the specification. The line doesn't really make sense so I deleted it.

bridgetkromhout (approved):
Thanks!

PR #1119, opened 2021-05-26
docs: fix chinese expression

JeffreyChu:
`集群` is better than `群集`, especially in computer world.


bridgetkromhout (approved):
Thank you, @JeffreyChu!

PR #1124, opened 2021-06-05
Fixed invalid configmap manifest in the tutorial.

AahanSingh:
The configmap manifest in the tutorial was invalid because the value being assigned to mug was boolean and not a string.
This caused issues with helm install without the --dry-run --debug flags because the manifest is invalid.
I have changed the manifest for configmap in the tutorial to use mug: "true" which no longer throws errors.

Signed-off-by: Aahan Singh <aahansingh95@gmail.com>


bridgetkromhout (approved):

bridgetkromhout:
Thank you, @AahanSingh!

PR #1133, opened 2021-06-16
Add example on using auto roll deployment checksum with library chart

nickjj:
Hi,

I'm new to using Helm and I really like how the docs guided you towards getting a checksum of a file to control automatically rolling out your deployments if a file changes, such as a config map or secret.

But when I transitioned to using library charts the approach in the docs no longer worked because it seems library charts aren't included in the `$.Template.BasePath` path.

This PR addresses this by adding an example that works with library charts.


bridgetkromhout (approved):
Thank you, @nickjj! Appreciated.

PR #1135, opened 2021-06-17
Add 3.6.x to version_skew doc

ecksun:
3.6.x was missing from the version_skew documentation.

I'm quite certain it is true but I haven't tested it.

I also don't know if there are other docs that needs updating


bridgetkromhout (approved):
Thanks! We missed updating this on the recent release. Appreciated.

PR #1140, opened 2021-06-21
fix: typo

KevinWu0904:
Signed-off-by: KevinWu0904 <KevinWu0904@outlook.com>


hickeyma (approved):
Checked with Google translate and fix makes sense. LGGTM, thanks @KevinWu0904

PR #1142, opened 2021-06-23
Update using_helm.md: Dot Escape

drewmarshburn:
Is the updated dot escape two backslash characters? Seems like only two backslash characters worked in a set for me (for v3.4.1). Found in a comment here: https://github.com/cetic/helm-nifi/issues/50#issuecomment-777882683.

### DIFF
--- a/content/en/docs/intro/using_helm.md
+++ b/content/en/docs/intro/using_helm.md
@@ -327,7 +327,7 @@ name: "value1,value2"
 
 Similarly, you can escape dot sequences as well, which may come in handy when
 charts use the `toYaml` function to parse annotations, labels and node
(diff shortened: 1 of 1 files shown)


bacongobbler:
Only one should be necessary as per `pkg/strvals`. Some terminal emulators may require you to add an extra backslash on the command line so it's not interpreted as a control character, as backslashes are used frequently as an escape character.

In other words this is correct, however depending on your environment, yes, you may require two backslashes to force the terminal to output one backslash.

See https://github.com/helm/helm/blob/main/pkg/strvals/parser.go#L398

(Runes require two backslashes as well, since they accept escape codes).

PR #1154, opened 2021-07-08
Updating the docs on dependencies

mattfarina:
- for when a repo isn't specified
- adding details on downloader plugins
- removing reference to the obsolete official repo


bridgetkromhout (approved):
Thanks!

PR #1161, opened 2021-07-31
[zh] Update Chinese doc

Zhanxueyou:


angellk (approved):
thanks @Zhanxueyou ! looks good

PR #1163, opened 2021-08-04
Add documentation for GitLab Chart Repo

diazjf:
Adds documentation providing information on using the GitLab Package Registry for Helm Charts.

### DIFF
--- a/content/en/docs/topics/chart_repository.md
+++ b/content/en/docs/topics/chart_repository.md
@@ -229,6 +229,10 @@ You can also use the
 [ChartMuseum](https://chartmuseum.com/docs/#using-with-local-filesystem-storage)
 server to host a chart repository from a local file system.
 
+### GitLab Package Registry
+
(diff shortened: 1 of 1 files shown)


bacongobbler (changes requested):
Hi @diazjf. Generally speaking third party services are listed here: https://helm.sh/docs/community/related/

Would you mind adding it there instead? Thanks.

diazjf:
@bacongobbler thanks for taking a look!

I looked into https://helm.sh/docs/community/related/ but I noticed those are mainly plugins and additional tools. What GitLab has introduced is a Helm Chart Repository which is listed in this commit along with other 3rd party repositories like:

- ChartMuseum
- JFrog Artifactory
- GitHub Pages
- Etc.

Therefore the location I'm proposing the change to is valid. Let me know your thoughts.

bacongobbler:
Ah yes... I just noticed we have links to several other cloud providers here. I guess my point is moot then.

bacongobbler (approved):

PR #1165, opened 2021-08-05
Updates on chain of trust/sigstore

bridgetkromhout:
A community member asked and so we discussed in today's weekly Helm developer call: what is this open line of inquiry around chain of trust? These doc updates are the result of that discussion.

- Helm is not currently looking at using the OSI model for signatures
- A relevant plugin sigstore exists, which can be linked in the docs


angellk (approved):
lgtm - thanks @bridgetkromhout

Issue #1157, opened 2021-07-15
chart archives not supported without repo


mattfarina:
In the docs at https://helm.sh/docs/topics/charts/#managing-dependencies-manually-via-the-charts-directory it says you can use a chart directory or archive for those in the charts directory without a repo being specified. This is an error. You can only have a directory and not a chart archive. The docs need to be corrected.

bridgetkromhout:
Resolved in https://github.com/helm/helm-www/pull/1168

PR #1168, opened 2021-08-07
Fix error about dependency chart format

alex-petrov-vt:
Closes issue #1157 

Signed-off-by: Alex Petrov <alex.petrov.vt@gmail.com>


bridgetkromhout (approved):
Thanks, @alex-petrov-vt!

PR #1175, opened 2021-08-19
[en] Clarify the absense of Sprig's OS functions

alexnoz:
Signed-off-by: Alex Nozdriukhin <alex-nozzz@mail.ru>

### DIFF
--- a/content/en/docs/howto/charts_tips_and_tricks.md
+++ b/content/en/docs/howto/charts_tips_and_tricks.md
@@ -15,7 +15,7 @@ resource files. While Go ships several built-in functions, we have added many
 others.
 
 First, we added all of the functions in the [Sprig
-library](https://masterminds.github.io/sprig/).
+library](https://masterminds.github.io/sprig/), except `env` and `expandenv`, for security reasons.
 
 We also added two special template functions: `include` and `required`. The
 `include` function allows you to bring in another template, and then pass the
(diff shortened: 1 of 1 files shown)


content/en/docs/howto/charts_tips_and_tricks.md
```diff
@@ -15,7 +15,7 @@ resource files. While Go ships several built-in functions, we have added many
 others.
 
 First, we added all of the functions in the [Sprig
-library](https://masterminds.github.io/sprig/).
+library](https://masterminds.github.io/sprig/), except `env` and `expandenv`, for security reasons.
```
bridgetkromhout, line 18: I see that these are still excluded. The reason in the past was due to security concerns around Tiller:
> First, we added almost all of the functions in the Sprig library. We removed two for security reasons: env and expandenv (which would have given chart authors access to Tiller’s environment)
https://v2.helm.sh/docs/developing_charts/#know-your-template-functions

Before updating this documentation, let's ensure that (even with Tiller gone in Helm v3) the reason for eliminating these two functions is still security-related.

alexnoz: I found a [PR](https://github.com/helm/helm/pull/5815) that would enable these functions, it was rejected because of security concerns, as follows from the discussion, for example see this [comment](https://github.com/helm/helm/pull/5815#issuecomment-499223593):
>This may still be considered a big security risk. By providing charts access to the environment, we could be exposing details about the running environment the user may not want the chart to have. The fact that the rendering engine was exposing Tiller's environment is a much larger security risk, however the same risk still applies even if Tiller was being run locally.

PR #1182, opened 2021-09-02
Add Flux to Helm docs under Community > Related Projects > Tools

scottrigby:
- Sort related projects alphabetically
- Fix YAML linting along the way
- Add Flux to Tools section


angellk (approved):
awesome - thanks @scottrigby !

PR #1191, opened 2021-09-08
docs: add post-renderer example

charles-woshicai:
Signed-off-by: charles-woshicai <583605554@qq.com>


content/en/docs/topics/advanced.md
```diff
@@ -80,6 +80,97 @@ type PostRenderer interface {
...
+		log.Fatal(err)
+	}
+	// Print post rendered manifests to console and it would be used by helm actions `install`, `test` and `template`.
+	fmt.Println(out)
+}
+```
+
+##### Run below commands to apply the custom post-renderer executor:
+```shell
+go build -o custom-postrender custom.go
```
bridgetkromhout, line 170: Earlier, you named the file differently:

> Here is `custome.go`

If these are intended to be the same, please name them consistently to help prevent confusion - thanks.

charles-woshicai: Sure, should both be `custom.go`

bridgetkromhout:
We need to do a tech review on this docs update to ensure its accuracy. Thanks for your patience, @charles-woshicai!

content/en/docs/topics/advanced.md
```diff
@@ -80,6 +80,97 @@ type PostRenderer interface {
 
 For more information on using the Go SDK, See the [Go SDK section](#go-sdk)
 
+#### Simple Example
```
joejulian, line 83: I feel like this should be at an H3 rather than an H4. H4 makes it look like it's part of the section on how to use the SDK, but this doesn't use the SDK.

joejulian (commented):
This is a good example of how to use the `post-renderer` flag and I have confirmed that the example works as intended.

content/en/docs/topics/advanced.md
hickeyma: Agree, lets make it a Header 3 instead i.e. `###`

content/en/docs/topics/advanced.md
```diff
@@ -80,6 +80,97 @@ type PostRenderer interface {
...
+
+Here is `custom.go`
```
hickeyma, line 85: can we change this sentence to be more descriptive.? Something like the following: This is a simple example of a custom post renderer which reads the Helm rendered manifests, adds a label to `deployment` and then outputs the modified manifests.

hickeyma (changes requested):
This is a really nice example @charles-woshicai, thanks. Just some small nits to fix and it should be good to go.

TerryHowe (changes requested):
I don't think this should be added to the docs because it will be hard to support code in docs that is not tested. If anything, this should be moved to sdkexamples

PR #1100, opened 2021-05-12
Improve rendering of commands for shell completion

zigarn:
Before:
image

After:
image

PR #1206, opened 2021-09-23
Improve rendering of commands for shell completion in translations

zigarn:
Follow-up helm/helm#9718
Renew #1100


bridgetkromhout (approved):
Thanks, @zigarn!

PR #1212, opened 2021-09-29
fix: wrong URL to mapkubeapis

morremeyer:
I was just debugging why I couldn’t install `mapkubeapis` when I noticed that I was trying to install @hickeyma's fork, not the upstream repo. This was due to a wrong link.

This PR fixes the link.


hickeyma (approved):
Thanks for catching this @morremeyer. FYI, it wasn't a fork originally but where the  tool was created and resided. Not so long ago, I transferred it to Helm org.

PR #1232, opened 2021-11-03
`and` and `or` functions can take many arguments

haines:
The docs currently state that the `and` and `or` functions take two arguments, but this isn't true - they are variadic.

I think a similar change is needed in Chinese and Korean, but unfortunately I don't speak either of those languages!

### DIFF
--- a/content/en/docs/chart_template_guide/function_list.md
+++ b/content/en/docs/chart_template_guide/function_list.md
@@ -34,7 +34,8 @@ Helm includes numerous logic and control flow functions including [and](#and),
 
 ### and
 
-Returns a boolean and of the two arguments.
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Thanks, @haines! We'll have a maintainer of the helm/helm codebase validate this update and then hopefully get it in. Appreciated.

bacongobbler (approved):
confirmed: https://go.dev/play/p/yTgXx35rQic

bridgetkromhout:
> I think a similar change is needed in Chinese and Korean

@haines The contributors for those languages pull in changes from the English docs from time to time, so this is a fine place to update it. Thanks!

haines:
Fantastic, thanks both!

PR #1249, opened 2021-12-19
Added note precreating namespace or repository for helm push

sabre1041:
Added a note for `helm push` to OCI registries which may require namespace and/or repository to be created beforehand.

Related to https://github.com/helm/helm/issues/10169


scottrigby (approved):
Thanks Andy! 👍 

follow-up from https://github.com/helm/helm/issues/10169#issuecomment-976973309

PR #1263, opened 2022-01-23
fix translation error about --dry-run code. in zh get start.md

zhujiaxin:
fix translation error about --dry-run code. 
更正了一些关于--dry-run的翻译错误"


zhujiaxin:
do not pass the check

PR #1271, opened 2022-01-29
Add helm-external-val to the list of helm plugins

kuuji:
Hey there 👋 

I recently wrote a plugin that allows to fetch values from external sources. Currently it supports fetching values from kubernetes configMaps and secrets  but I'm planning on adding support for more in the future.
I have examples on how to use it with argocd in the Readme.

Adding it here to get more visibility in case the helm community is interested.


bridgetkromhout (approved):
Thanks, @kuuji!

PR #1273, opened 2022-02-03
fix grammar in chart_best_practices/values.md

toddmacintyre:
This PR corrects a small grammar mistake found here: https://v2.helm.sh/docs/chart_best_practices/#consider-how-users-will-use-your-values .


hickeyma (approved):
LGTM, thanks for the fix @toddmacintyre

Issue #1288, opened 2022-03-08
Update Hugo Install Instructions or Refactor for Hugo v0.93.x Compatibility


Timothy-Dement:
## Problem Description

I was going through the [README](https://github.com/helm/helm-www/blob/main/README.md) to get set up for a minor documentation fix and hit a snag after [`brew install hugo`](https://github.com/helm/helm-www/blob/50c0f424e3cdf8363c456f31689ecc4b90646ce5/README.md?plain=1#L12).

The current version of `hugo` is `v0.93.3` (as of `08 Mar 2022`) and that version is installed via `brew`:

- https://formulae.brew.sh/formula/hugo

If `hugo serve` is run in the `helm-www` repository root, several errors occur:

```
> $ hugo serve
--------------
Start building sites … 
hugo v0.93.3+extended darwin/amd64 BuildDate=unknown
WARN 2022/03/08 14:45:27 found no layout file for "HTML" for kind "section": You should create a template file which matches Hugo Layouts Lookup Rules for this combination.
ERROR 2022/03/08 14:45:27 render of "page" failed: execute of template failed: template: blog/single.html:5:7: executing "main" at <partial "nav.html" .>: error calling partial: execute of template failed: template: partials/nav.html:30:11: executing "partials/nav.html" at <partial "nav-i18n.html" .>: error calling partial: "/Users/tmdement/Code/helm-www/themes/helm/layouts/partials/nav-i18n.html:21:68": execute of template failed: template: partials/nav-i18n.html:21:68: executing "partials/nav-i18n.html" at <$translation.URL>: can't evaluate field URL in type page.Page
ERROR 2022/03/08 14:45:27 render of "page" failed: execute of template failed: template: blog/single.html:5:7: executing "main" at <partial "nav.html" .>: error calling partial: execute of template failed: template: partials/nav.html:30:11: executing "partials/nav.html" at <partial "nav-i18n.html" .>: error calling partial: "/Users/tmdement/Code/helm-www/themes/helm/layouts/partials/nav-i18n.html:21:68": execute of template failed: template: partials/nav-i18n.html:21:68: executing "partials/nav-i18n.html" at <$translation.URL>: can't evaluate field URL in type page.Page
[...]

PR #1290, opened 2022-03-11
[1288] update site template to work with newer hugo versions

flynnduism:
This was a pretty small edit to resolve the `hugo` build error documented in #1288. 

Here I replace the depricated `.URL` code and the site builds without any issue on the latest Hugo version (v0.94.0 ).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ rootfs/svc
 .yarnclean
 _site/*
 _site/
+.hugo_build.lock
 .ncurc.yml
 .publish/*
 .publish/
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,7 +3,7 @@
   publish = "app"
 
 [build.environment]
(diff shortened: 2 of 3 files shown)


Timothy-Dement (approved):
LGTM - one question, the empty `.hugo_build.lock` file isn't in `main` but seems to be generated with each `hugo serve` command - does it matter whether or not it's included?

hickeyma (changes requested):
It looks good apart from the `.hugo_build.lock` file, as @Timothy-Dement raised. Should this is added to `.gitignore`?  This is mentioned [here](https://discourse.gohugo.io/t/what-is-the-hugo-build-lock-file/35417).

flynnduism:
Yes I will add the lock file to the `.gitignore`

hickeyma (approved):
LGTM, thanks @flynnduism

PR #1310, opened 2022-05-10
oci chart is now supported in Captain

zhhray:
update document contents, oci chart is now supported in Captain.


bridgetkromhout:
@scottrigby Since these updates are to a _blog post_ not to docs specifically, please decide how you want to handle this - thanks.

PR #1315, opened 2022-05-25
Docs: explain conditional behaviour of 'with'

plumdog:
I have seen (and, in fact, written) plenty of templates that use this behaviour, but couldn't find it documented within helm.

I did, however, find it documented here https://pkg.go.dev/text/template#hdr-Actions.


bridgetkromhout (approved):
Helpful clarification - thank you!

PR #1324, opened 2022-06-15
added next and previous to the blog page for better user experience

AakashRaj20:


AakashRaj20:
@bacongobbler Hey I have raised the PR for this issue, can you review it and tell if its all right or I have to do some more change. Thank you

karenhchu:
Hi there! Thanks for your contribution. My thoughts here are that we should keep the arrows as is. If we're aiming to be as inclusive as possible, using universal symbols where we can is the best way to go as opposed to using words of a specific language. Thoughts @bacongobbler?

bridgetkromhout:
Hi, @AakashRaj20 - As Karen mentioned, we are trying to minimize the English that will appear in the navigation controls, because the Helm website works in multiple languages. We will close this PR without merging it, but please do add other PRs if desired; we'd be happy to have you contribute. Thanks.

PR #1337, opened 2022-07-20
add install helm with dnf/yum method on fedora.

aysabzevar:
Signed-off-by: Ali Yousefi Sabzevar <aysabzevar@gmail.com>


bridgetkromhout (approved):
Thanks!

PR #1353, opened 2022-09-09
Provide better explanations of dot context

rehevkor5:
- See https://github.com/helm/helm/issues/11232 and https://github.com/helm/helm/issues/7915 for examples of confusion the current explanation/wording has caused.
- Replaced inconsistent and unexplained terminology: "root context", "root scope", "top-level scope", "parent scope".
- Go templates/gomplate, on which Helm relies, uses the words "dot" and "context", not "scope". The word "scope" is better used to explain when variables go out of scope, not to explain the value of the context variables `.` and `$`.
- Fixed inaccurate or misleading statements such as: $ "does not change", and $ "will always point to the root context".
- This sentence: "Recall that `.` is a reference to _the current scope_." is not useful - this was not explained elsewhere
- Removed vague sentences that don't add meaning, such as "Values are namespaced, but namespaces are pruned."

Signed-off-by: Shannon Carey <rehevkor5@gmail.com>

### DIFF
--- a/content/en/docs/chart_template_guide/named_templates.md
+++ b/content/en/docs/chart_template_guide/named_templates.md
@@ -156,9 +156,9 @@ templates, it is best to name your templates with _chart specific names_. A
 popular naming convention is to prefix each defined template with the name of
 the chart: `{{ define "mychart.labels" }}`.
 
-## Setting the scope of a template
+## Setting the context of a template
 
-In the template we defined above, we did not use any objects. We just used
+In the template we defined above, we did not use any objects from the dot context `.`. We just used
 functions. Let's modify our defined template to include the chart name and chart
 version:
 
@@ -173,7 +173,7 @@ version:
 {{- end }}
 ```
 
-If we render this, we will get an error like this:
+If we render this with `template` as above, we will get an error like this:
 
 ```console
 $ helm install --dry-run moldy-jaguar ./mychart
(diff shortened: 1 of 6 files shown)


miles-w-3:
This is a great set of changes, I didn't know that `with` could act as a conditional operator

rehevkor5:
Anything I can do to move this forward?

content/en/docs/chart_template_guide/named_templates.md
```diff
@@ -190,17 +190,18 @@ metadata:
...
 the template like this:
 
 ```yaml
 {{- template "mychart.labels" }}
 ```
 
-No scope was passed in, so within the template we cannot access anything in `.`.
-This is easy enough to fix, though. We simply pass a scope to the template:
+No pipeline value was provided to its second parameter, so within the template `.` is empty.
+`$.` will therefore also be empty.
```
VannTen, line 203: I think it should be `$` without the dot here.

rehevkor5: I see what you mean. I guess I got caught by the inconsistency that `.` is analogous to `$` but `.Foo` is analogous to `$.Foo` (extra dot appears from thin air) instead of `$Foo`. Something about `.` is magical I guess.

```
% echo "{{ . }}" | gomplate
map[]
% echo "{{ $. }}" | gomplate
16:41:50 ERR  error="template: -:1: unexpected <.> in operand"
```

Fixing...

yurrriq:
I wonder if it would be better to eschew the redundant parts (e.g. `if`, `with`, `range`) in favor of a link to https://pkg.go.dev/text/template, which explains the actions nicely.

rehevkor5:
> I wonder if it would be better to eschew the redundant parts (e.g. if, with, range) in favor of a link to https://pkg.go.dev/text/template, which explains the actions nicely.

You might be right, though I didn't want to make this change any more difficult to accept than it probably already is.

If such a alteration would make this PR more likely to be accepted, I would be happy to do it.

rehevkor5:
Maybe someone like @yxxhero can review?

SaintRyoh:
The current helm documentation on this is very misleading, and will easily make you believe that structures [ like this](https://github.com/helm/helm/issues/7915) are supposed to work in helm templates:
`
{{ with .Values.something }}
  {{ template "mychart.template" . }}     # dot for callee = .Values.something
{{ end }}

{{- define "mychart.template" -}}
name: {{ include "mychart.fullname" $ }}  # here, $ == . (the passed argument)
var:  {{ .otherVar }}
{{- end -}}
`
When is this going to be merged?

rehevkor5:
Rebased again, not sure how to get attention on it. @yxxhero or @bacongobbler ?

rehevkor5:
Fun fun.

TerryHowe:
Sorry, docs were converted to docusaurus and you have conflicts now.

rehevkor5:
gg

PR #1359, opened 2022-09-30
Add "zot" as OCI-based registry which supports helm charts.

rchincha:
Signed-off-by: Ramkumar Chinchani <rchincha@cisco.com>


rchincha:
Just FYI, https://github.com/project-zot/zot
^ zot is a OCI registry project sponsored by Cisco Systems, Inc.

https://github.com/project-zot/zot/blob/main/test/blackbox/pushpull.bats#L122

angellk (approved):
lgtm - thanks @rchincha

PR #1366, opened 2022-10-15
Fix typo in plugin YAML example

emranbm:
The "name" field of the sample YAML was "last" and should replace "keybase" in the doc as its a typo.

Signed-off-by: Emran BatmanGhelich <emran.bm@gmail.com>


emranbm:
@hickeyma The point on the `platformCommand` makes sense; but, in a few lines below that YAML snippet, it's stated that:
> _`name` should match the directory name._ In our example above, that means the
plugin with `name: keybase` should be contained in a directory named `keybase`.  

Based on your comment, shouldn't it be `name: last` in the above sentence? Am I missing something?

hickeyma:
@emranbm Yes, `keybase` should be replaced with `last` everywhere. I think it was a copy/paste from a previous Keybase example.

emranbm:
> @emranbm Yes, `keybase` should be replaced with `last` everywhere. I think it was a copy/paste from a previous Keybase example.

Sorry for the late response. Done renaming all `keybase`s to `last` in the English documentation of plugins. Please review it again. Thanks.

hickeyma (approved):
LGTM, thanks for fixing this @emranbm

PR #1403, opened 2022-11-29
Added details on library chart differences from application charts

miles-w-3:
Signed-off-by: Miles Wilson <wilson.mil@icloud.com>

In general, there is confusion around the use case for accessing files between subcharts and parent charts, exemplified through https://github.com/helm/helm/issues/1892.

The goal of this commit is to shed more light on the benefits of using library charts. As far as I can tell, these differences are not explicit the official docs, and many people seem to think that there is very little benefit in using library charts beyond extracting code to be used in multiple charts. Through experimentation, I have found the two peices of functionality documented in this commit very useful:

- Accessing the entire .Values dict from the parent chart, not just the slice with the subchart header
- Being able to access the parent chart's files from a library chart.

In particular, the second piece has allowed for extremely versatile chart development, and I think more advertisement of the chart features could lead to increased adoption of library charts and less confusion around file access. I have attempted to update the library charts page itself as well as add caveats for absolute statements that seem to only apply to application charts (presumably leftover from before library charts existed). Thanks for your time, I'm happy to add/edit details as needed.

### DIFF
--- a/content/en/docs/chart_template_guide/accessing_files.md
+++ b/content/en/docs/chart_template_guide/accessing_files.md
@@ -21,6 +21,8 @@ this works:
   security reasons.
   - Files in `templates/` cannot be accessed.
   - Files excluded using `.helmignore` cannot be accessed.
+  - Files outside of a helm application [subchart]({{< ref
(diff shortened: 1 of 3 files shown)


angellk (approved):
This looks great - appreciate you taking the time to add clarification @miles-w-3 !

PR #1419, opened 2023-01-20
Include push as a function that works with OCI registries

### LINKED ISSUE #1418: Registry Topic list of functions Incomplete for OCI
In https://helm.sh/docs/topics/registries/#other-subcommands

It states:
> Support for the oci:// protocol is also available in various other subcommands. Here is a complete list:

That list fails to include the `push` command which has many examples just above. So either `push` should be in the list or the list is not "complete" but both cannot be true.

Given the other examples showing push in the docs, I am assuming push should function with OCI registries ( I cannot test it on my side ).

mcdan:
Adds "push" as a valid function for OCI registries. Fixes #1418


TerryHowe (approved):
/lgtm

PR #1423, opened 2023-01-27
Add howto reuse chart multiple times in dependency doc

nymo:
### Why this PR

I recently searched for a solution how to reuse a chart multiple times. Unfortunately the documentation about how to add dependencies had no such information.
I found something about it in the Chart.yml keyword description.
In my opinion this should also be mentioned in the documention about how to add dependencies because this is the first place where people start looking.


angellk (approved):
thanks for adding this @nymo - lgtm

PR #1438, opened 2023-03-12
docs: Remove gofish from index samples

craigbox:
In learning what a gofish was, I see that it's now deprecated. Instructions are not in the docs; this commit updates the home page to match.


craigbox:
Confirmed renders OK in deploy preview.

PR #1439, opened 2023-03-13
Fix wrong korean translation

minkyu97:
Fix a translation that could be interpreted as the opposite.


angellk (approved):
good catch - thanks @minkyu97

PR #1443, opened 2023-03-21
add a simple example for merge function

monegim:
Add a simple example for a better understanding the `merge` function. The values also have been taken from the `mergeOverwrite` example to be compared the results.


angellk (approved):
thanks for the helpful example @monegim - lgtm

PR #1459, opened 2023-05-04
functions_list: reorder merge example

chtitux:
https://github.com/helm/helm-www/pull/1443 inserted the example in the middle of the text.


angellk:
@chtitux the original makes more sense to me versus reordering - could you please explain your thinking?

PR #1463, opened 2023-05-06
fix cheat sheet documentation typo in helm list

ram191:
**Issue**

There is a typo in the cheat sheet documentation. The command helm list --all and helm list --all-namespaces is written with only a leading single dash.

image


angellk (approved):
nice catch - thanks @ram191

ram191:
Glad to help!

PR #1467, opened 2023-05-10
Updating for Helm 3.12

mattfarina:


angellk (approved):
awesome - thanks @mattfarina

PR #1483, opened 2023-07-03
Make cheatsheet appear properly on side navigation and next page

anilkumarmyla:
Currently cheatsheet doesn't show up in the sidebar navigation and "next" page title shows as empty for "Using Helm" page, hopefully this fixes that issue


angellk (approved):
thanks for fixing the side nav @anilkumarmyla ! lgtm

PR #1491, opened 2023-07-26
docs: fix commands in helm completion bash

Stringls:
Hello. I was setting up helm completion in bash today following [these docs](https://helm.sh/docs/helm/helm_completion_bash/) and noticed that BASH commands are using output redirection to `/etc/` and `/usr` directories which requires `sudo` permissions, but the redirection of the output is not **performed** by `sudo` so it leads to the following error.

```bash
bash: /etc/bash_completion.d/helm: Permission denied
```

In that case I propose to update the docs to use `sudo tee` command that redirects output to the file and prints output to the terminal at the same time.
**NOTE**: If you don't like the output in the terminal I can add `> /dev/null` at the end of the command to redirect output of `tee` command


flynnduism:
Thank you @Stringls! LGTM

flynnduism (approved):

PR #1498, opened 2023-08-24
Adds dt to the list of plugins

mpermar:
Adding a link to a new plugin that we just have open sourced.


yxxhero (approved):
thanks so much.

PR #1499, opened 2023-08-28
fix(doc): helm list go sdk doc

Ithrael:
related issue: https://github.com/helm/helm/issues/12269

[helm sdk simple-example](https://helm.sh/docs/topics/advanced/#simple-example)
This document is incorrect, according to this document, helm sdk can't list all charts. I would like to make a modification in the helm project so that helm sdk does not need to invoke client.SetStateMask(), but it is very difficult. Because helm sdk can filter through client.All=True, and can also filter through client.StateMask=129. I can't meet these two situations with very small modifications. In addition, I noticed that in the source code of helm, after setting client.All=True, it also needs to invoke client.SetStateMask(). https://github.com/helm/helm/blob/main/cmd/helm/list.go#L237 So, I decided to modify the documentation and add client.SetStateMask() to the examples provided in the helm sdk documentation.


Ithrael:
@gjenkins8 @yxxhero please review the PR

Ithrael:
@BigStupideer @angellk please review the PR, thank you.

TerryHowe:
I don't think this change applies anymore, thanks just the same. Reopen if I missed something.

PR #1507, opened 2023-09-17
fix: typo in Flow Control document

atobaum:
fix type: end -> and


Benbb96 (commented):
`{{ end }}` seems to be good for me. It's the closing tag after `{{ if ... }}`

angellk:
agree with @Benbb96 ! closing for now

PR #1508, opened 2023-09-22
Add external project HULL to list

gre9ory:
Hello,

I wanted to propose a mentioning of this open source project here. 

It is built in form a Helm library chart and has been around for more than two years now. Many tests exist to back every releases stability and it is used in production.  

Not sure if it should better go under the plugins sections but these seem to be (almost) exclusively actual helm plugins there?

Thanks!


angellk (approved):
lgtm - this is a good place to add HULL - thanks @gre9ory

gre9ory:
thank you very much @angellk!

PR #1517, opened 2023-10-19
add section about the `required` function

knulpi:
Adding a section about the `required` function, that is for now only mentioned in the [tips and tricks](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) section


angellk (approved):
lgtm - thanks @knulpi

PR #1519, opened 2023-11-10
Simplify hooks example

SuperSandro2000:


content/en/docs/topics/charts_hooks.md
```diff
@@ -127,8 +127,8 @@ metadata:
     app.kubernetes.io/version: {{ .Chart.AppVersion }}
     helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
SuperSandro2000, line 128: Technically all those labels are also not required and removing them would allow to focus on the important parts and make this way easier to grasp and understand :)

PR #1523, opened 2023-11-18
docs(chart_best_practices): callout error with YAML comments

OneCricketeer:
Using `# {{ .Values.undefinedVar }}` in a template can cause unexpected errors for those unfamiliar with the `{{/* */}}` block syntax.


angellk (approved):
thanks for the addition @OneCricketeer - lgtm

PR #1535, opened 2023-12-09
Fix web site scrolling on mobile browsers

### LINKED ISSUE #741: Improve site on mobile
It’s difficult to select the hamburger menu icon in the upper right hand corner on mobile with the BLM banner in the way on the main page. Would it be possible to add a button to either hide the message or raise/overlay the hamburger menu icon so that the menu icon and the BLM messages are both displayed?

I’m trying to view on site on an iPhone 7.

### LINKED ISSUE #1393: Website is unusable on mobile due to scroll container
Using Chrome on Android, scroll is unusable. The page seems to be wrapped in a container that has its own scroll bar.

### LINKED ISSUE #1464: Vertical scrolling over Helm docs.
The docs over at https://helm.sh/docs/intro/cheatsheet/ has a transient error so it doesn’t allow for scrolling up or down on IOS Safari. It’s pretty bad.

Same goes for Firefox on IOS..

lightster:
Scrolling pages on the Helm web site is difficult on mobile devices due to nested scrolling containers. This changeset removes the nested scrolling containers.

The fix to remove the nested scrolling containers caused pages to start scrolling horizontally because the content of the pages have negative margins, which makes the page wider than 100%. The rest of this changeset addresses the negative margins by removing them and/or offsetting them in each element that was causing the page to horizontal scroll.

Fixes #741
Fixes #1393
Fixes #1464

### DIFF
--- a/themes/helm/assets/sass/bulma/sass/components/level.sass
+++ b/themes/helm/assets/sass/bulma/sass/components/level.sass
@@ -1,5 +1,9 @@
 $level-item-spacing: ($block-spacing / 2) !default
 
+.columns.level-columns
+  margin-left: 0
+  margin-right: 0
+
 .level
   @extend %block
   align-items: center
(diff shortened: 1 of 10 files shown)


scottrigby:
This has been addressed a different way, and should now be fixed since the docusaurus migration. Please open new issues for specific OS/device bugs if you still experience them.

PR #1536, opened 2023-12-12
Depreacted installation using snap packages

mangin:
Hey there,

I would like to deprecate installation using snap package.

Reason:
Because Snapcrafters stopped to release new versions of helm. And it could create some problems for new comers that would pick up this way of installation of helm. Because they would use pretty old version of helm.

Proof:
The actual version of helm [3.12.2]. The latest snap version of helm is [3.10.1]. The latest release was made Oct 12, 2022. So the package wasn't updated more than for 1 year.

Links:
1. https://github.com/helm/helm/releases/tag/v3.13.2
2. https://github.com/helm/helm/releases/tag/v3.10.1
3. https://snapcraft.io/helm


mattfarina:
Sorry it took a little bit to get to this. I just checked things out and they have the latest Helm version. It seems they are now keeping up to date. For now, let's close this and wait until there is a problem again.

Thanks for alerting us to this.

PR #1538, opened 2023-12-13
fix: add permission to write repo & upgrade action to v1.6.0

warjiang:
1. `helm/chart-releaser-action` need write permission to run it correctly otherwise it will report permission error 
image

2. upgrade `helm/chart-releaser-action` to v1.6.0


angellk (approved):
lgtm - thanks @warjiang

PR #1542, opened 2024-01-02
Chore: fix Home overview section to show a transparent background color on IOS

### LINKED ISSUE #1537: Home overview section is showing a transparent background color on IOS
Running `helm.sh` on a Chrome browser in an iPhone shows a transparent background. 
You will notice this mostly in the overview section beneath the hero area, and the footer section.
IMG_8547

cc: @scottrigby

thisisobate:
Fixes #1537

### DIFF
--- a/themes/helm/assets/sass/docs-responsive.scss
+++ b/themes/helm/assets/sass/docs-responsive.scss
@@ -155,9 +155,16 @@
 
 // mobile and tablet combined
 @media only screen and (max-width: 1023px) {
+  .home-overview-wrap {
+    background-image: linear-gradient(180deg, darken($navy, 5%) 0%, $navy 100%);
+    background-blend-mode: normal;
+  }
+
   section.footer-links {
     padding-left: 10vw;
     padding-right: 10vw;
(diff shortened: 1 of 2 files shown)


thisisobate:
I just realized the issue isn't with the linear gradient not supported on MacOS.
The real problem is that `background-blend-mode` css property is not supported on IOS. See here: https://stackoverflow.com/questions/35452263/css-a-background-blend-mode-fallback-for-iphone-safari

As a fallback, I'm considering disabling the background image and set to a background color instead. This only applies to mobile and tablet displays across all browsers. The idea is to prioritize usability over aesthetics.

How did I arrive at my current solution?

I considered setting the background not to repeat as the [stackoverflow article](https://stackoverflow.com/questions/35452263/css-a-background-blend-mode-fallback-for-iphone-safari) suggests but that didn't work in my case. Another option I considered was to set an overlay over the background image and tweak the opacity a bit but that did not give me a very desirable result.

Hope this helps!

mattfarina (approved):

PR #1552, opened 2024-02-07
charts_tips_and_tricks.md: remove misleading wording about quotes

SpecLad:
Quotation marks around `helm.sh/resource-policy` are not required, and moreover, the usual YAML style is to not include quotes in keys unless necessary.

Remove the quotes and the claim that they are required.

Fixes helm/helm#12555.


angellk (approved):
lgtm - thanks @SpecLad

PR #1555, opened 2024-02-16
Translation in French of Helm commands

BenjaminFourmaux:
Translation of all Helm commands in French


angellk (approved):
lgtm @BenjaminFourmaux - merci!

PR #1558, opened 2024-03-07
fix: remove double equal signs

fty4:
Hi there,

in the [`Debugging Templates`](https://helm.sh/docs/chart_template_guide/debugging/) docs  is a double equal signs in the docs for using `--dry-run=server`.
Only one should be used here.


angellk (approved):
nice catch - thanks @fty4

PR #1567, opened 2024-03-27
describe mergeOverwrite with nil field in src (#3)

michaelruigrok:
If any keys in the mergeOverwrite function's src dict are nil, they will also be set to nil in the dst dict

### DIFF
--- a/content/en/docs/chart_template_guide/function_list.md
+++ b/content/en/docs/chart_template_guide/function_list.md
@@ -1445,10 +1445,12 @@ dst:
   default: default
   overwrite: me
   key: true
+  nullify: me
 
 src:
   overwrite: overwritten
   key: false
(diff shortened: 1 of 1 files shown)


content/en/docs/chart_template_guide/function_list.md
```diff
@@ -1458,6 +1460,7 @@ newdict:
   default: default
   overwrite: overwritten
   key: false
+  nullify: nil # AKA null
 ```
```
TerryHowe, line 1464: I think adding this to the main example is just going to cause confusion for most people that are not concerned with this case. I think adding some prose down further along the lines of "Overriding a dictionary with a `nil` value does not delete the key value pair, it will not result in a dictionary with a key to a `nil` value"

but IMNAM

PR #1570, opened 2024-04-01
Added --generate-name flag for command

TheLoGgI:
Command resolves to an error, saying 'must either provide a name or specify --generate-name', adding the flags fixes that.
Alternative adding "rude-cardinal" (same name as previous guides) `helm install rude-cardinal --dry-run --debug mychart`


angellk (approved):
thank you for adding the clarification @TheLoGgI - lgtm

PR #1576, opened 2024-04-20
Added examples for using `Chart Starter Packs` and a few explanations.

ksemele:
I added more context for using starters because the current explanation is a little bit confusing.


angellk (approved):
Lgtm - thanks for this @ksemele - helpful to have additional examples!

PR #1583, opened 2024-05-22
charts_tips_and_tricks.md - quote password in case of special characters

elwin013:
Add quoting (using `| quote`) for password in imagePullSecret helper and removing additional (artificial) quote 
characters as those are added by `quote` func.

This change allows to use more complicated password with e.g. quotes, backslashes and other special characters.


TerryHowe (approved):
/lgtm

elwin013:
I've updated the new files with same changes (should be applied now to all files).

@TerryHowe wdyt? :slightly_smiling_face:

TerryHowe (approved):
/lgtm

PR #1591, opened 2024-06-12
added generic Dockerfile and .dockerignore

tomerle03:
added `Dockerfile` and `.dockerignore` for docker build in disconnected environments and self serving.


robertsirc:
Just so i understand you want to airgap the helm website?

mattfarina (changes requested):
What's the use case for adding this?

PR #1605, opened 2024-07-11
Fix small typo

adrianmoisey:
I think this is supposed to be a "tenet"


hickeyma (approved):
LGTM, thanks @adrianmoisey

PR #1623, opened 2024-09-14
Add Aptakube to "Related Projects and Documentation"

goenning:
Aptakube has supported Helm for some time now, and a user recently suggested adding it here.

image


angellk (approved):
lgtm - thanks @goenning !

PR #1629, opened 2024-10-16
feature: add documentation on OCI install with digest

TerryHowe:
Documentation for OCI digest install feature.

Related https://github.com/helm/helm/pull/12690


TerryHowe:
Available as of 3.18

PR #1631, opened 2024-10-18
Added ProGet to list of Chart Repositories in Docs

whatatripp:
This seems to be one of those circumstances an issue might be overkill; hopefully the details provided here will suffice in this case.

Basically I just wanted to improve the docs by pointing to another solution for hosting private chart repositories. 

Labels/tags would be `docs`  and `size/XS`


TerryHowe (approved):
/lgtm

but IANAM

PR #1633, opened 2024-10-22
Update en.toml

karenhchu:
adding KubeCon NA '24 to upcoming events section


karenhchu:
looks like this won't actually update the section; need to figure out updating the whole website another time

PR #1634, opened 2024-10-24
Remove Weaveworks and Bloodorange, and unused logo images

scottrigby:
From today's helm dev meting.

A policy will be created about criteria for logos on the Helm website footer.

This PR addresses cleanup that can happen even before the new policy is outlined:
- Weaveworks and Bloodorange can be removed from website footer because the companies no longer exist. 
- Also removed Codecademy logo, as it was removed from footer years ago when I moved from that company, and the image was never cleaned up from the logowall directory.


karenhchu (approved):
LGTM! Thanks for addressing this so quickly.

PR #1637, opened 2024-11-05
Improve helm lint documentation clarity

myk0la-b:
Clarified helm lint --with-subcharts option documentation

Please note https://github.com/helm/helm/issues/12027 -- abovementioned unclear part caused many people to struggle

_Please help with other languages documentatiton_

### DIFF
--- a/content/en/docs/helm/helm_lint.md
+++ b/content/en/docs/helm/helm_lint.md
@@ -35,7 +35,7 @@ helm lint PATH [flags]
       --skip-schema-validation    if set, disables JSON schema validation
       --strict                    fail on lint warnings
   -f, --values strings            specify values in a YAML file or a URL (can specify multiple)
-      --with-subcharts            lint dependent charts
+      --with-subcharts            lint subcharts as independent charts (note that without this option, the chart will be validated as a whole, including subcharts as its part)
 ```
 
 ### Options inherited from parent commands


blackoutnet (approved):

astellingwerf:
To me, the documentation is still unclear: I don't understand what this option does, especially since the linting behavior is so different from `helm template`. It seems impossible to lint the subscharts if they use template provided by the superchart.

myk0la-b:
@astellingwerf, so the thing I am trying to emphasize there is that when you run `helm lint` **without** the `--with-subcharts` option, your subchart will be linted as part of the superchart. i.e. it exactly does
> lint the subscharts if they use template provided by the superchart

But on the other hand, when the `--with-subcharts` option **is used**, the subchart would be additionally linted as an independent entity.

content/en/docs/helm/helm_lint.md
```diff
@@ -35,7 +35,7 @@ helm lint PATH [flags]
       --skip-schema-validation    if set, disables JSON schema validation
       --strict                    fail on lint warnings
   -f, --values strings            specify values in a YAML file or a URL (can specify multiple)
-      --with-subcharts            lint dependent charts
+      --with-subcharts            lint subcharts as independent charts (note that without this option, the chart will be validated as a whole, including subcharts as its part)
```
mattfarina, line 38: This document is generated from Helm itself. So are all the files in this directory.

It needs to be changed at https://github.com/helm/helm/blob/ecc4adee692333629dbe6343fbcda58f8643b0ca/cmd/helm/lint.go#L149.

Please update it in the other location.

mattfarina (changes requested):

content/en/docs/helm/helm_lint.md
andrask: Additionally, the wording is still hazy. Maybe something like this could be better:
```suggestion
      --with-subcharts            lint subcharts separately from the parent chart
```
To be honest, this option's name is already misleading. But that's much harder to change.

PR #1653, opened 2024-11-14
fix: kindIs function documentation

### LINKED ISSUE #1532: kindIs returns false when documentation says it should return true
Opening this issue following the reply on https://github.com/helm/helm/issues/12084 (https://github.com/helm/helm/issues/12084#issuecomment-1558177496)

## Description

`{{ if kindIs "int" 123 }}` returns false when [documentation says it should return true](https://docs.helm.sh/docs/chart_template_guide/function_list/#reflection-functions).

> This appears to be a documentation issue. The kind of a yaml/json parsed number is float64 with the library that's used. Please file an issue in [helm-www](https://github.com/helm/helm-www/issues).

Behavior details can be found in mentioned issue.

TerryHowe:
The `kindIs` function does not work as documented because the underlying library used for parsing stores number values as `float64`

Closes: https://github.com/helm/helm-www/issues/1532


andrask:
Instead of highlighting one example that's inconsistent with the expectations, I think it would be better to add details about how such errors can be debugged. Or even better, warn to not use the kindIs at all, and instead throw errors when a conversion / assertion fails. Failing fast with any kind of configuration errors is the best strategy anyway.

TerryHowe:
> Instead of highlighting one example that's inconsistent with the expectations, I think it would be better to add details about how such errors can be debugged. 

Right, but this PR is about an inconsistency not an error.

Issue #1399, opened 2022-11-23
`$` in the context of named templates: 


thernstig:
Reference: https://github.com/helm/helm/issues/8702

https://helm.sh/docs/chart_template_guide/variables/ says:
> However, there is one variable that is always global - $ - this variable will always point to the root context.

This is misleading, as in effect, the "global scope" is limited to the scope passed to the named template. As @bacongobbler says:
> "global scope" is a bit of a misnomer in this case. In actuality, the dollar sign operator refers to the root scope. In a named template, the root scope changes into the form of the input.

It would be good if both https://helm.sh/docs/chart_template_guide/variables/ is updated to mention this, as well as mentioning this important fact in https://helm.sh/docs/chart_template_guide/named_templates/#setting-the-scope-of-a-template.

th0ger:
Spent hours on this misleading docs. Please fix :-)

scottrigby:
fixed in #1662

PR #1662, opened 2024-12-04
clarify global scope meaning for named templates

lgerard-pass:
I just got bitten by the fact that you cannot use the `$` sign to access the root context when using a named template.

This should solve #1399

### DIFF
--- a/content/en/docs/chart_template_guide/named_templates.md
+++ b/content/en/docs/chart_template_guide/named_templates.md
@@ -219,7 +219,7 @@ metadata:
 
 Note that we pass `.` at the end of the `template` call. We could just as easily
 pass `.Values` or `.Values.favorite` or whatever scope we want. But what we want
-is the top-level scope. Note that in the context of the named template `$` will refer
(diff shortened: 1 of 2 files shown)


andrask (approved):
LGTM

content/en/docs/chart_template_guide/named_templates.md
```diff
@@ -219,7 +219,8 @@ metadata:
 
 Note that we pass `.` at the end of the `template` call. We could just as easily
 pass `.Values` or `.Values.favorite` or whatever scope we want. But what we want
-is the top-level scope.
+is the top-level scope. Note that in the context of the named template `$` will refer
+to the scope you passed in and not some global scope.
```
TerryHowe, line 223:
```suggestion
is the top-level scope. In the context of the named template `$` will refer
to the scope you passed in and not some global scope.
```
A nit, but a Note in a Note was a bit much

TerryHowe (approved):
/lgtm

but IANAM

PR #1667, opened 2024-12-21
Adding redirect for helm/v4 to the code

mattfarina:
Helm uses custom URLs to get to the code. In this case helm.sh/helm/v4 should point to the main branch for helm so that people who try to use it with go modules are able to.

This change adds that redirect to the default language which means it should work for everyone.


yxxhero (approved):
LGMT

sabre1041 (approved):
LGTM

karenhchu:
LGTM!

PR #1670, opened 2024-12-23
fix: suggest using registry over repository

TerryHowe:
During my Helm booth duty at Kubecon in SLC a user had not considered using a registry to store charts. It seems like that should be more clear as a possibility.


scottrigby (approved):
lgtm

karenhchu:
Thanks @TerryHowe!

PR #1679, opened 2025-01-29
docs: add a warning about default value for condition

PascalBourdier:
this warning seems important because the default value is usually `false`

### DIFF
--- a/content/en/docs/chart_best_practices/dependencies.md
+++ b/content/en/docs/chart_best_practices/dependencies.md
@@ -58,6 +58,9 @@ is in a sub-directory of the `charts` folder with the name being the same as the
 
 Conditions or tags should be added to any dependencies that _are optional_.
 
+> [!IMPORTANT]
+> By default a `condition` is `true`.
+
 The preferred form of a condition is:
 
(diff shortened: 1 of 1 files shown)


obourgain (approved):
aussi un warning sur la forme inline du yml qui ne fonctionnait pas ?
```
mmp-batch-multitenant.enabled: false
```

TerryHowe (commented):
I am not a maintainer, but I think the stress on this is out of character for the web site.

content/en/docs/chart_best_practices/dependencies.md
```diff
@@ -58,6 +58,9 @@ is in a sub-directory of the `charts` folder with the name being the same as the
 
 Conditions or tags should be added to any dependencies that _are optional_.
 
+> [!IMPORTANT]
+> By default a `condition` is `true`.
+
```
TerryHowe, line 63:
```suggestion
Conditions or tags should be added to any dependencies that _are optional_. Note that by default a `condition` is `true`.
```

TerryHowe (approved):
/lgtm

PR #1696, opened 2025-02-22
Format template example

dnskr:
The PR formats template example by adding indentation to improve readability.

### DIFF
--- a/content/de/docs/howto/charts_tips_and_tricks.md
+++ b/content/de/docs/howto/charts_tips_and_tricks.md
@@ -168,9 +168,9 @@ Wir definieren dann unsere Hilfsvorlage wie folgt:
 
 ```
 {{- define "imagePullSecret" }}
-{{- with .Values.imageCredentials }}
-{{- printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"%s\",\"auth\":\"%s\"}}}" .registry .username .password .email (printf "%s:%s" .username .password | b64enc) | b64enc }}
-{{- end }}
+  {{- with .Values.imageCredentials }}
+    {{- printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"%s\",\"auth\":\"%s\"}}}" .registry .username .password .email (printf "%s:%s" .username .password | b64enc) | b64enc }}
+  {{- end }}
 {{- end }}
 ```
 
--- a/content/en/docs/howto/charts_tips_and_tricks.md
+++ b/content/en/docs/howto/charts_tips_and_tricks.md
@@ -170,9 +170,9 @@ imageCredentials:
 We then define our helper template as follows:
 ```
 {{- define "imagePullSecret" }}
-{{- with .Values.imageCredentials }}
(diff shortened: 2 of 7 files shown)


TerryHowe (approved):
/lgtm

TerryHowe:
Before
Screenshot 2025-03-09 at 8 00 01 AM

TerryHowe:
After:
Screenshot 2025-03-09 at 7 59 52 AM

dnskr:
@yxxhero Could you please review the PR?

dnskr:
@scottrigby Could you please review and merge the PR?

dnskr:
Force pushed to address conflicts and to add changes to `docs/howto/charts_tips_and_tricks.md` file.

scottrigby:
@dnskr first of all thanks for your contribution! Happy to see more. However, I think this specific change should be reverted.

I know these .tpl files look awful not indented, but it's tricky - conditionals and printf for example are fine, bue we don't want to encourage folks to indent in general. For example, take the common labels helper block generated by `helm create foo`:

```
{{/*
Common labels
*/}}
{{- define "foo.labels" -}}
helm.sh/chart: {{ include "foo.chart" . }}
{{ include "foo.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}
```

First of all, how would you indent something like this? From your example I'd think like this below. It doesn't look as pretty as the simple example you changed above.

But no matter how you do indent it, any of the text is indented, it will be malformed YAML:

```
{{/*
Common labels
*/}}
{{- define "foo.labels" -}}
    helm.sh/chart: {{ include "foo.chart" . }}
    {{ include "foo.selectorLabels" . }}
    {{- if .Chart.AppVersion }}
        app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
    {{- end }}
        app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}
```

```console
% helm template foo foo
Error: YAML parse error on foo/templates/deployment.yaml: error converting YAML to JSON: yaml: line 9: mapping values are not allowed in this context

Use --debug flag to render out invalid YAML
```

This is why these definitions all on one line.

dnskr:
@scottrigby Thank you for the comment. I see the provided example with common labels as an exception to the good style rule due to limitations of the templating engine. I believe we should still encourage indentation for templates in YAML (an indentation-based format) to avoid cases like one of the following real-life example:
```yaml
{{- if and .Values.statsd.enabled }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-statsd
  labels:
    tier: airflow
    component: config
    release: {{ .Release.Name }}
    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
    heritage: {{ .Release.Service }}
{{- with .Values.labels }}
{{ toYaml . | indent 4 }}
{{- end }}
data:
  mappings.yml: |-
{{- if .Values.statsd.overrideMappings }}
    mappings:
{{ toYaml .Values.statsd.overrideMappings | indent 6 }}
{{- else }}
{{ .Files.Get "files/statsd-mappings.yml" | indent 4 }}
{{- if .Values.statsd.extraMappings }}
{{ toYaml .Values.statsd.extraMappings | indent 6 }}
{{- end }}
{{- end }}
{{- end }}
```

scottrigby:
@dnskr you're mixing examples - this PR was indenting helper definitions, which should not be indented. Your example just above is for templates, which you're right should be indented (except `nindent` is better in general than `indent` for templates). See https://github.com/helm/helm-www/pull/1895#issuecomment-3525990076

Issue #1020, opened 2021-02-16
Docs: Helm template function and "must" versions of functions


thernstig:
https://helm.sh/docs/chart_template_guide/function_list/ has many `must` versions of functions, such as `mustRegexMatch` vs. `regexMatch`. The [docs](https://helm.sh/docs/chart_template_guide/function_list/#regexmatch-mustregexmatch) reads:

> `regexMatch` panics if there is a problem and `mustRegexMatch` returns an error to the template engine if there is a problem.

As many of the function has these `must` equivalents it might be a good idea to explain somewhere in the docs what the above means, potentially with an example. I do not find it clear what difference I will see as a user if these functions fail. I "assume" I should rather use the `must` versions, but I am not sure.

mattfarina:
@thernstig That's a good idea. I've moved the issue to the repository that holds the docs. It has less traffic and will hopefully be more likely to get attention.

thernstig:
@mattfarina Thanks. Is there any chance you would know if `must` is preferred?

bacongobbler:
Depends on the situation, really. Some cases would prefer an error when the regular expression does not match with the given pattern. Other times it's okay for no matches to be returned. Both will return an error if the regular expression fails to compile.

thernstig:
I see. I would have guessed that in the above case, if a user inserts an incorrect `.Values.currentDate` it would have been nicest to use the `mustToDate` still. The reason being, otherwise the Helm is installed with an unexpected value for the user.

But I agree it is situational still, and many might not agree with me there.

johnlinvc:
One issue we got is when using `mustRegexMatch` as an assertion, it'll generate extra true (from the return value) in the rendered result.
We need to add extra steps to get rid of the return value.
For example

```
{{- $val := "val" -}}
{{- mustRegexMatch "a" $val -}}
{{- print $val -}} # outputs trueval 
```

A hack to fix this is discard the result with
```
{{- $val := "val" -}}
{{- mustRegexMatch "a" $val | toString | trunc 0 -}}
{{- print $val -}} # outputs val 
```

Does anyone have a better way to discard the result without using `if` ?

bstrdsmkr:
Since this is still open with a question and I found it trying to understand the must variants, I think a common convention has evolved to assign the value to "_" and just forget it ever happened :smile: 
```go
{{- $_ := regexMatch "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?" $key }}
[...]

PR #1698, opened 2025-03-01
fix(30216): Add more docs for must version of functions

DerekTBrown:
## What does this PR do?
- Extend documents to address https://github.com/helm/helm/issues/30216 and https://github.com/helm/helm-www/issues/1020.

### DIFF
--- a/content/en/docs/chart_template_guide/function_list.md
+++ b/content/en/docs/chart_template_guide/function_list.md
@@ -857,18 +857,38 @@ Helm includes the following regular expression functions: [regexFind
 (mustRegexReplaceAllLiteral)](#regexreplaceallliteral-mustregexreplaceallliteral),
 [regexSplit (mustRegexSplit)](#regexsplit-mustregexsplit).
 
-### regexMatch, mustRegexMatch
+### regexMatch
 
-Returns true if the input string contains any match of the regular expression.
+Returns `true` if the input string contains any match of the regular expression.
 
 ```
-regexMatch "^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}$" "test@acme.com"
+{{- if regexMatch "^abc$" "abc" -}}
+valid: true
+{{- end -}}
+```
+The above produces `valid: true`.
+
+`regexMatch` will ignore regex compilation errors, and simply return `false` ([ref](https://github.com/Masterminds/sprig/blob/master/regex.go#L7))
+
+```
+{{- if regexMatch "[" "abc" -}}
+valid: true
+{{- end -}}
+```
+The above produces nothing.
+
+### mustRegexMatch
+
+Similar to `regexMatch`, but will return an error to the templating engine if the regex fails to compile:
+
+```
+{{- if mustRegexMatch "[" "abc" -}}
+valid: true
+{{- end -}}
 ```
 
-The above produces `true`
+The above produces `error: "error parsing regexp: missing closing ]`.
 
-`regexMatch` panics if there is a problem and `mustRegexMatch` returns an error
-to the template engine if there is a problem.
 
 ### regexFindAll, mustRegexFindAll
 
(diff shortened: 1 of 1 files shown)


content/en/docs/chart_template_guide/function_list.md
```diff
@@ -857,18 +857,38 @@ Helm includes the following regular expression functions: [regexFind
 (mustRegexReplaceAllLiteral)](#regexreplaceallliteral-mustregexreplaceallliteral),
 [regexSplit (mustRegexSplit)](#regexsplit-mustregexsplit).
 
-### regexMatch, mustRegexMatch
+### regexMatch
```
TerryHowe, line 860: There are links above and this change breaks that. I don't think you should change the headers or add the `### mustRegexMatch` header here.

content/en/docs/chart_template_guide/function_list.md
```diff
@@ -857,18 +857,38 @@ Helm includes the following regular expression functions: [regexFind
...
+```
+{{- if regexMatch "[" "abc" -}}
+valid: true
+{{- end -}}
+```
+The above produces nothing.
+
+### mustRegexMatch
+
+Similar to `regexMatch`, but will return an error to the templating engine if the regex fails to compile:
```
TerryHowe, line 882:
```suggestion
In the event of regex compilation errors, `regexMatch` will simply return `false`.
Use `mustRegexMatch` to fail in the event of regex compilation errors.
```

content/en/docs/chart_template_guide/function_list.md
```diff
@@ -885,7 +905,7 @@ The above produces `[2 4 6 8]`
 `regexFindAll` panics if there is a problem and `mustRegexFindAll` returns an
 error to the template engine if there is a problem.
 
-### regexFind, mustRegexFind
+### regexFind
```
TerryHowe, line 908:
```suggestion
### regexFind, mustRegexFind
```

content/en/docs/chart_template_guide/function_list.md
```diff
@@ -895,8 +915,15 @@ regexFind "[a-zA-Z][1-9]" "abcd1234"
 
 The above produces `d1`
 
-`regexFind` panics if there is a problem and `mustRegexFind` returns an error to
-the template engine if there is a problem.
+`regexFind` will result in a panic if the regex fails to compile ([ref](https://github.com/Masterminds/sprig/blob/master/regex.go#L30)), whereas
+`mustRegexFind` will return an error to the templating engine if the regex fails to compile ([ref](https://github.com/Masterminds/sprig/blob/master/regex.go#L34)).
```
TerryHowe, line 919: I think these references are going to be brittle
```suggestion
The `regexFind` function will result in a panic if the regex fails to compile, whereas
`mustRegexFind` will return an error to the templating engine if the regex fails to compile.
```

content/en/docs/chart_template_guide/function_list.md
```diff
@@ -895,8 +915,15 @@ regexFind "[a-zA-Z][1-9]" "abcd1234"
...
+
+Note that neither function will panic/return an error if the input string does not match the regex. Both functions return an empty string in this case:
```
TerryHowe, line 921: This project doesn't strictly use semantic line breaks, but I think it helps.
```suggestion
Note that neither function will panic/return an error if the input string does not match the regex.
Both functions return an empty string in this case:
```

TerryHowe (changes requested):
I am not a maintainer, but there are some changes that need to happen here, thanks!

TerryHowe (changes requested):
```
9:09:12 PM: docs/chart_template_guide/function_list/index.html
9:09:12 PM:   hash does not exist --- docs/chart_template_guide/function_list/index.html --> #regexfind-mustregexfind
9:09:12 PM:   hash does not exist --- docs/chart_template_guide/function_list/index.html --> #regexmatch-mustregexmatch
```

PR #1699, opened 2025-03-03
Update args that will be consumed by the root helm command

artemmikhalitsin:
The flag parsing section of the plugins guide does not contain an up-to-date list of global helm args.
```
      --burst-limit int               missing in documentation and currently passed to plugins
      --kube-apiserver string    missing in documentation
      --kube-as-group stringArray    missing in documentation
      --kube-as-user string             missing in documentation
      --kube-ca-file string             missing in documentation
      --kube-insecure-skip-tls-verify   missing in documentation
      --kube-tls-server-name string     missing in documentation
      --kube-token string              missing in documentation
      --qps float32                    missing in documentation and currently passed to plugins
```

First, this PR re-orders the flags so that they appear in alphabetical order - the same order as `helm --help`

This PR adds the missing global args to the plugin documentation. I made a best effort to also update the other languages.
For Ukrainian (uk) - I can mostly vouch. One difference I'd like to point out is I used the plural `Ці перетворюються` for -kube-as-user, as this is a stringArray parameter.
For Korean, I simply reused the same format as other strings. If someone could help with using a plural form, feel free to add a suggestion.

Some flags in this PR are currently also passed to plugins. For this reason, I provide a pr in helm/helm:
https://github.com/helm/helm/pull/30614

### DIFF
--- a/content/en/docs/topics/plugins.md
+++ b/content/en/docs/topics/plugins.md
@@ -334,14 +334,22 @@ will be set as the `KUBECONFIG` variable
 
 When executing a plugin, Helm will parse global flags for its own use. None of
 these flags are passed on to the plugin.
-
+- `--burst-limit`: This is converted to `$HELM_BURST_LIMIT`
 - `--debug`: If this is specified, `$HELM_DEBUG` is set to `1`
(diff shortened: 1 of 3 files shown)


yxxhero (approved):
LGTM

PR #1722, opened 2025-05-09
add playground options to Helm

abhi-bhatra:
## Summary
This PR adds references to Free Playground across key tutorial sections of the Helm documentation, allowing users to try Helm without local installation. Similar to how Kubernetes documentation offers interactive learning environments, these additions allow new Helm users to quickly experiment with commands in a browser-based environment.

## Motivation
- Many new users want to try Helm before installing it locally
- Setting up a Kubernetes cluster locally can be challenging for beginners
- Interactive learning environments accelerate the learning process

## Impact
This change will:
- Make Helm more accessible to new users
- Reduce the learning curve barrier
- Provide a consistent learning path similar to Kubernetes documentation

Closes: #1719


yxxhero:
@abhi-bhatra please fix DCO issue.

abhi-bhatra:
Thanks for the help @yxxhero. Fixed the issue

TerryHowe (commented):
I'm not a maintainer here, but
* I think this should just be in the getting started.
* While I agree those are great resources, linking external resources can be problematic

abhi-bhatra:
So, what do you suggest @TerryHowe ? What changes I can do here: 
- keeping this only in getting-started ?
- And what else ?

TerryHowe:
Wait for some more eyes on this.

TerryHowe:
Yes

PR #1734, opened 2025-06-28
Document the annotation added to manifests passed to the post-renderer

carloslima:
The annotation was added as part of helm/helm#13154

### DIFF
--- a/docs/topics/advanced.mdx
+++ b/docs/topics/advanced.mdx
@@ -36,7 +36,7 @@ non-0 exit code in the event of a failure. This is the only "API" between the
 two components. It allows for great flexibility in what you can do with your
 post-render process.
 
-A post renderer can be used with `install`, `upgrade`, and `template`. To use a
+A post-renderer can be used with `install`, `upgrade`, and `template`. To use a
 post-renderer, use the `--post-renderer` flag with a path to the renderer
 executable you wish to use:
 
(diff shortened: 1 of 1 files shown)


docs/topics/advanced.mdx
```diff
@@ -48,9 +48,19 @@ If you wish to use multiple post-renderers, call all of them in a script or
 together in whatever binary tool you have built. In bash, this would be as
 simple as `renderer1 | renderer2 | renderer3`.
 
-You can see an example of using `kustomize` as a post renderer
+You can see an example of using `kustomize` as a post-renderer
 [here](https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render).
 
+#### Filenames
+Each manifest passed to the post-renderer includes a temporary annotation:
```
TerryHowe, line 58: It kind of sounds like each manifest is past to the post-render and if they have the annotation, the file name is not changed.

carloslima: I struggled with the wording a little bit. Do you have a suggestion that would make it clearer?

In most cases, the post-renderer can safely ignore the annotation. It will then be passed back to Helm, which uses it to restore its internal state. However, the post-renderer may modify the annotation if needed. It can also add manifests and use the annotation to indicate which file they originated from, if that information is useful.

TerryHowe:
```suggestion
Each manifest passed to the post-renderer may include a temporary annotation:
```

carloslima: Every manifest passed into the post-renderer is guaranteed to have the annotation. However, the manifests output by the post-renderer may or may not include the annotation.

In other words, the manifests read from STDIN by the post-renderer will have the annotation, but the manifests written to STDOUT may not. This can happen if the post-renderer deletes the annotation, modifies it, or adds new manifests without it.

scottrigby:
needs rebase after the docusaurus migration, if you want to pick back up on this.

carloslima:
> needs rebase after the docusaurus migration, if you want to pick back up on this.

Done

TerryHowe (approved):
/lgtm

PR #1743, opened 2025-07-10
Clarify that the `with` action has an `else` clause

lindhe:
Before this PR, there was no indication in Helm's docs that the `with` action has an `else` clause. This change clarifies that `else` exists for the `with` action.

### DIFF
--- a/content/en/docs/chart_template_guide/control_structures.md
+++ b/content/en/docs/chart_template_guide/control_structures.md
@@ -259,11 +259,13 @@ The next control structure to look at is the `with` action. This controls
 variable scoping. Recall that `.` is a reference to _the current scope_. So
 `.Values` tells the template to find the `Values` object in the current scope.
 
-The syntax for `with` is similar to a simple `if` statement:
+The syntax for `with` is similar to the `if`/`else` block:
 
 ```
 {{ with PIPELINE }}
-  # restricted scope
+  # restricted scope, if PIPELINE is true
+{{ else }}
+  # restricted scope, if PIPELINE is false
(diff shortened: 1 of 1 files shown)


TerryHowe (commented):
Seems reasonable, but DCO issue

content/en/docs/chart_template_guide/control_structures.md
```diff
@@ -259,11 +259,13 @@ The next control structure to look at is the `with` action. This controls
...
 
-The syntax for `with` is similar to a simple `if` statement:
+The syntax for `with` is similar to the `if`/`else` block:
 
 ```
 {{ with PIPELINE }}
-  # restricted scope
+  # restricted scope, if PIPELINE is true
+{{ else }}
+  # restricted scope, if PIPELINE is false
```
jkroepke, line 268:
```
{{with pipeline}} T1 {{else}} T0 {{end}}
	If the value of the pipeline is empty, dot is unaffected and T0
	is executed; otherwise, dot is set to the value of the pipeline
	and T1 is executed.
```

Reading the official docs, the else scope is not restricted.

lindhe: Oh, that's so obvious when you say it, thanks for catching it! 🤦 Fix incoming!

lindhe: I've pushed a new commit now with both proper signoff and changed this to not be restricted. Mark this as resolved if you are happy with the change. Otherwise, I'll change it again according to your suggestions.

PR #1674, opened 2025-01-15
Update website for Helm Release v3.17.0

scottrigby:
Part of release checklist

PR #1759, opened 2025-09-01
Point to the right last major version 3.18.x

benoittgt:
Screenshot 2025-09-01 at 16 46 13

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -45,7 +45,7 @@ jsFiles = [
 ]
 
 [[params.versions]]
-version = "v3.17.0"
+version = "v3.18.0"
 patch = "stable"
 url = "https://helm.sh/docs"
 primary = true


yxxhero (approved):

scottrigby (changes requested):
whenever we bump the current version, we also need to also bump the next version. Example for this current PR:
```diff
- [params.nextversion]
- version = "v3.18.0"
+ [params.nextversion]
+ version = "v3.19.0"
```

And for future ref, other files. I realize this is a PR to catch up to a previously missed bump, so just noting for future that you can see this as an example of how it should look for MINOR releases: https://github.com/helm/helm-www/pull/1674

TerryHowe:
In theory, we should also bump the version in `sdkexamples/go.mod`

benoittgt:
Sorry, I probably rushed my patch. It should be better now.

config.toml
```diff
@@ -55,7 +55,7 @@ version = "v2.17.0"
 url = "https://v2.helm.sh/docs"
 
 [params.nextversion]
-version = "v3.18.0"
+version = "v3.19.0"
 date = "May 14, 2025"
```
TerryHowe:
```suggestion
date = "October 1, 2025"
```

I don't know the planned date for 3.19

benoittgt: Good point. I took the date from https://github.com/helm/helm/releases/tag/v3.18.6

benoittgt:
Not needed anymore with https://github.com/helm/helm-www/pull/1772

PR #1762, opened 2025-09-05
Prevent hiding "stable" text on version switcher

benoittgt:
While looking at a doc version for https://github.com/helm/helm-www/pull/1759, I noticed that the "stable" text was hidden.

Screenshot 2025-09-05 at 11 08 47
Screenshot 2025-09-05 at 11 08 43

This patch changes the orientation of the drop-down to avoid hiding this important information.

Screenshot 2025-09-05 at 11 14 00
Screenshot 2025-09-05 at 11 13 41


sabre1041 (approved):
LGTM

PR #1767, opened 2025-09-08
Automate document update process

TerryHowe:
Add an automated periodic update job that checks for new Helm releases and updates the documentation accordingly.

## How it works

The update job runs daily at 2am Pacific (9am UTC) and performs the following steps:

1. **Check for new releases**: Fetches the latest releases from the [Helm GitHub repository](https://github.com/helm/helm/releases)
2. **Filter releases**: Ignores pre-release versions and focuses on stable releases
3. **Update versions**: Updates version numbers in `config.toml` for new minor releases of v3 and v2
4. **Update dependencies**: Updates the Go module dependencies in `sdkexamples/go.mod` for v3 updates
5. **Regenerate documentation**: Runs the Helm CLI documentation generator to update command references
6. **Create pull request**: If changes are made, creates a pull request with the updates


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,73 @@
...
+  schedule:
+    # Run daily at 2am Pacific (9am UTC)
+    - cron: '0 9 * * *'
+  workflow_dispatch: # Allow manual trigger
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
```
gjenkins8, line 24: don't we want to limit the execution here to main branch?

gjenkins8: (if this happens already, this could just be I've not remembered how GitHub actions work)

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: |
+          if git diff --quiet; then
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+          else
+            echo "has_changes=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: steps.changes.outputs.has_changes == 'true'
+        uses: peter-evans/create-pull-request@v7
```
gjenkins8: we should pin version via release hash

TerryHowe:
I agree that "no need to check for Helm 2 updates - there are no more", but that can stand as a placeholder for v4 when that become available. I don't want to remove the capability to update two releases.

.github/workflows/update-docs.yml
TerryHowe: This job will only run on main.  The actions/checkout just checks out the repo on the default branch.

The github action is trigged by the schedule and manually (workflow_dispatch). It is not trigged on a push to any branch.

.github/workflows/update-docs.yml
TerryHowe: Done.

TerryHowe:
Notes in _italic_ about what is done and what is not or partial:

Split the functionality:

- [x] I'm thinking there should be two scripts: one to poll for newer Helm versions, and another to update the docs in all the places we need for a new version _(Instead of this, I've changed the script to work on one version at a time which simplifies it. Currently it is less than 200 lines.)_
- [x] the first can be on a cron, and can trigger the second passing the version as an arg to workflow_dispatch
- [x] when the workflow_dispatch is manually triggered, the user can pass the version as the argument

More simplifications:

- [x] no need to check for Helm 2 updates - there are no more _(Consider it a placeholder for when v4 is available)_
- [x] we can remove docker call from the script - the workflow will run on ubuntu already _(this makes manual runs difficult)
- [x] the readme for this script may be too verbose, especially given the details may change with the new docs site. 

Other details:

- [X] all files need a newline at EOF
- [X] the make target name should probably be update-docs-version or something more specific than update-docs

TerryHowe:
Updated for new site. Seems to work.

TerryHowe:
We can do something else for this. I think rather than automate this aspect of a release, we should automate the release process via a prompt to perform the operations.

PR #1781, opened 2025-09-16
Added further mention of helm with snap

ppbrown:
With currently live documentation, Ubuntu users may not realize that they dont have to install the new repo: they can just install the snap versions.

Change doc to give reference in "Ubuntu" section to the snap section, and also mention "k8s helm"


TerryHowe:
conflicts

PR #1783, opened 2025-09-21
add note about ed25519 key issue

jessesimpson36:
In the following helm issue: https://github.com/helm/helm/issues/31181

we discovered that the presence of an ed25519 key anywhere in your keyring will cause all package signing to fail even if you configure helm to sign with an RSA key (in a keyring that happens to have an ed25519 key within it).  This is a side-effect of using an outdated openpgp library that is only receiving security updates.

This change will notify users that when exporting their keys, they should try to ONLY export the keys that they plan on using for the helm signature, and not all of their keys.


jessesimpson36:
2025-09-21-174820_grim

jessesimpson36:
This is no longer relevant as newer versions of helm have patched this bug.

PR #1788, opened 2025-09-28
docs / fix helm template cli flags 

TimaevAlexandr:
drop extra flags from `helm template` cli docs


TerryHowe:
Reopen if I missed something

PR #1789, opened 2025-10-03
Redirect to Helm 4 Release Party Sponsorship Prospectus

scottrigby:


sabre1041 (commented):
It looks fine. Maybe change the redirect to be `/v4-launch-party-sponsorship` in case we want to have another event in the future. 

Otherwise good with me 👍

scottrigby:
> It looks fine. Maybe change the redirect to be `/v4-launch-party-sponsorship` in case we want to have another event in the future.

good thinking. done

gjenkins8 (approved):

Issue #1730, opened 2025-06-02
HIP-0024: Evaluate static site generators and migrate the docs


paigecalvert:
Issue to track implementing "Evaluate static site generators and migrate the docs" from HIP: https://github.com/helm/community/blob/main/hips/hip-0024.md

Evaluate, select, and migrate the Helm docs to a modern, open-source static site generator (like [Docusaurus](https://docusaurus.io/) or [MkDocs](https://www.mkdocs.org/)) that would improve ease-of-contribution and site functionality.

paigecalvert:
Monday 06/23/25: Updated the [Helm Docs SSG Evaluation](https://docs.google.com/document/d/1Z-fHsGsZCyEEBrgN_IYhJSLehBYz-Kn8Pbv7jovFLZg/edit?tab=t.0#heading=h.9gvwwd6gr0xz) doc with my eval notes and recommendations. Shared with the Helm dev team here: https://kubernetes.slack.com/archives/C51E88VDG/p1750697074043249

The TLDR is that I'd lean towards one of the following options:
* Astro with Starlight integration: This one wasn't in our original list of SSGs, but I gave it a shot after Patrice from the CNCF tech docs team recommended it. Overall great; the only downsides were related to somewhat limited customization around the versioning functionality.
* Hugo with Docsy theme: This would be the ol' reliable option that has been the recommended SSG for CNCF projects for some time. It meets the Helm docs requirements and has the added benefit of lots of expertise/experience on the CNCF tech docs team. I didn't find it to be the easiest to work with of the bunch and I don't think it has the most modern look-and-feel, but it checks all the boxes (after some customization).

TerryHowe:
I don't have a vote here, but if I did, I'd vote for Hugo/Docsy.

From what you are evaluating, I've used Hugo, MkDocs and Docusaurus.

scottrigby:
Now that we have our first helm 4 pre-release out, I think we should hop on this.

Hugo would be fine.

My vote however would be for Docusauraus, since it solves some of the documentation problems really well, version handling experience for end users is really nice - and not hard for us to maintain. The only potential downside of Docusaurous Paige found online was possible performance issues at build time, which - in my opinion - are not an issue for us. We don't build that often, and we have finite content. The end user experience is fast and modern.

paigecalvert:
Closed by #1850

PR #1799, opened 2025-10-15
[HIP-0024] Rework CSS and JS and new homepage layout for Docusaurus

scottrigby:
This PR is part of [HIP-0024] tracking issue:
- https://github.com/helm/helm-www/issues/1730

Builds upon and should be merged after (in draft until then):
- https://github.com/helm/helm-www/pull/1791

This PR includes:
- Sitewide styles using Docusaurus Infima variables
- Homepage blocks as components
- Separate styles per component
- Update home page layout for clearer design
- Fix custom CSS, including: boat transition, reducing wave wrapper height,
  stacked solution for header content to allow scaling for different viewports
- Simplify custom JS, favoring docusaurus build-ins
- Hero block height calculator as client module
- Add architectural decisions doc to clarify decisions so far and enable
  easier contributions and maintenance
- Add blog styles and customize docs breadcrumbs

Co-authored-by: Paige Calvert <eura.calvert@gmail.com>
Signed-off-by: Scott Rigby <scott@r6by.com>

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -66,9 +66,6 @@ const config = {
             type: ["rss", "atom"],
             xslt: true,
           },
-          // Please change this to your repo.
-          // Remove this to remove the "edit this page" links.
-          editUrl: "https://github.com/helm/helm-www/blob/main/",
           // Useful options to enforce blogging best practices
           onInlineTags: "warn",
           onInlineAuthors: "warn",


scottrigby:
> * When there's no right side page toc, the margins of the main article go all the way to the right side of the viewport, which makes it a little harder to read with longer lines. This also happens on the landing page in that the "Read more" links are really far to the right of the screen. Not a big deal, but again just seems less polished/intentional when it's sort of stretched out

Fixed :+1:


> * Having an "Edit this page" link seems like something we don't want for blogs. Makes the blogs feel a little less polished/done to me, and also not sure if we want to encourage PRs

This is an interesting point I've never thought about. I agree docs - are something we generally want contributions on, but perhaps less so for blog posts published by specific authors. Removed :+1:

PR #1802, opened 2025-10-16
[HIP-0024] Add i18n for homepage, site elements, v3 docs, blogs

paigecalvert:
This PR is part of [HIP-0024] tracking issue:

- https://github.com/helm/helm-www/issues/1730

This PR:
- Updates the `i18n` key in `docusaurus.config.js` to add each locale and to configure the locale dropdown in the navbar
- Adds an `/i18n` directory with subdirs for each locale (done with the Docusaurus `write-translations` command)
- Updates each homepage component to add `<Translate>` tags around user-facing text. This leverages the Docusaurus translate API to add each string that's wrapped in `<Translate>` to a `code.json file` for the given locale after running `write-translations`
- Adds translations for the homepage in `i18n/[locale]/code.json file` for each locale (used Cursor/Claude, based on existing  `i18n/[locale].toml` files that were used by Hugo)
- Adds translations for the navbar in `i18n/[locale]/docusaurus-theme-classic/navbar.json` for all locales
- Adds translations for the footer  in `i18n/[locale]/docusaurus-theme-classic/footer.json` for all locales
- Moves existing translated v3 docs content for each locale to the appropriate i18n directory
- Moves existing translated blogs to the appropriate i18n directory
- Add translations for the sidebar categories to the `i18n/[locale]/docusaurus-plugin-content-docs/[docs-version].json` file
- Add translations to the `i18n/[locale]/docusaurus-plugin-content-blog/options.json` files
- Add `MARKDOWN_LINKS.md` file with guidance for writing markdown links that won't break across locales
- Removes `i18n/[locale].toml` files that were used by Hugo
- Updates `README-ko` to match the updates made to the English `README.md`. Used Cursor/Claude.


paigecalvert:
To test:

With `yarn`
```bash
yarn && yarn build 
yarn serve
```
Or with `npm`
```bash
npm install && npm run build
npm run serve
```

This builds all locales so you can preview the full site. Otherwise, you can do a local preview per locale (npm run start -- --locale es) or a local build per locale (npm run build -- --locale es).

scottrigby (changes requested):
Looking good! 

We need more translations, of course, but this bring over what we have and then some.

A few observations that should be updated:
- The `4.0.0-alpha.1 🚧` version label in the navbar versions dropdown only appears in English - all other languages say "Next".
- I like that proper nouns are left in English, such as "Kubernetes" and "Helm"
  - You may want to note this in the description above, and on our documentation page / CONTRIBUTING.md for criteria for helping with translations
  - I did notice that "Charts" is sometimes left in English, and sometimes not (see v2 docs sidebar, the 4th link down is "Charts" and it's sometimes translated)
  - This appears to be inconsistent on the live site, so we can't use that to make the call
  - Question - do we want to translate "Charts" or leave in English? Where do we draw the line? For example should helm "Releases" also be English, or is it fine to translate "chart" and "release"? I think this one should be made into a separate issue and raised at the weekly Helm dev meeting.
  - This appears to be inconsistent on the live site, so we can't use that to make the call
  - I would say let's not let that hold up this PR, since it's at least as good as what we have live now. Perhaps for this PR we should just choose one and go with it, then update later based on community feedback.

scottrigby:
Also, thinking about deploying - it takes a long time to build from scratch for a production build/serve (understandably, there are many languages with many files). We may want to make an issue for speeding up automated builds using caching from the previous build, since it seems to speed it up significantly even when there are translation changes.

paigecalvert:
> Also, thinking about deploying - it takes a long time to build from scratch for a production build/serve (understandably, there are many languages with many files). We may want to make an issue for speeding up automated builds using caching from the previous build, since it seems to speed it up significantly even when there are translation changes.

Issue added! https://github.com/helm/helm-www/issues/1811

PR #1820, opened 2025-10-20
[HIP-0024] Add blog author breadcrumbs

scottrigby:


scottrigby:
> It looks like the breadcrumbs themselves aren't getting translated.

Fixed:
- For the word "Authors" in breadcrumbs, we use the existing `theme.blog.authorsList.pageTitle`
- For the word "Blog" in breadcrumbs, here is a new `theme.blog.list.pageTitle`

I added these for Greek (el) to test, since Spanish, French etc use the same word for Blog as English.

test with `yarn start --locale el`

PR #1821, opened 2025-10-20
[HIP-0024] Fix waves height with announcement bar

scottrigby:
Before this, when the announcement bar is present, the waves/boat are pushed below the viewport when the announcement bar is present.


paigecalvert (approved):
LGTM! (Announcement bar + boat/waves both fully visible in the viewport)
Screenshot 2025-10-20 at 9 05 55 AM

PR #1818, opened 2025-10-19
[pl] initialize polish localization

dkarczmarski:
this PR is re-opened again in the [hip-0024-helm4-site-upgrade](https://github.com/helm/helm-www/tree/hip-0024-helm4-site-upgrade) branch as [scottrigby](https://github.com/scottrigby) asked in this comment:
https://github.com/helm/helm-www/pull/1806#pullrequestreview-3354594834 .

---

This PR initializes the Polish localization .
The implementation follows the official guidelines at https://helm.sh/docs/community/localization/ .
The changes are small and split into two clear commits .

PR #1824, opened 2025-10-20
exclude content/pl directory from typos check

dkarczmarski:
[pl] exclude content/pl directory from typos check

initialization of the content/pl directory is in an incubation phase : #1818 .

however, any build will fail without this update to `.types.toml` .

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -32,6 +32,8 @@ extend-exclude = [
   "i18n/",
   # Portuguese content - these are not English typos
   "content/pt/**",
+  # Polish content
+  "content/pl/**",
   # Russian content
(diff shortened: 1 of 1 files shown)


dkarczmarski:
> Do you want to include this in #1818 ?

@TerryHowe 
it would be good if I include this in #1818 ,
however this file ( `.typos.toml` ) exists in the main branch , but does not exist in the  hip-0024-helm4-site-upgrade branch yet .

EDIT:
... or I can cherry-pick entire file `.typos.toml` from main branch and add my changes

TerryHowe:
Ah, makes sense, thanks!

PR #1833, opened 2025-10-24
[HIP-0024] Update the netlify build command and env vars

paigecalvert:


scottrigby:
note I will change `onBrokenLinks` back to `throw` in a PR that fixes the broken links

PR #1859, opened 2025-10-30
Fix hero height calculator timing issue on mobile

scottrigby:
initialization using a reliable delay-based approach.


scottrigby:
I'm going to close this for now, I can not reproduce the earlier issues I was seeing.

PR #1865, opened 2025-11-02
fix style in intro/CheatSheet.md

dkarczmarski:
fix style in intro/CheatSheet.md

unify this document by:
1. all comments begin with a capital letter and end with a period "." .
I think that the period "." at the end is not necessary for single-sentence comment,
however for multi-sentence comments it seems justified.
so for consistency all comments end with a period.
2. remove unnecessary symbol, such as ":"
3. write comments in a form of order - where it is possible

### DIFF
--- a/versioned_docs/version-3/intro/CheatSheet.md
+++ b/versioned_docs/version-3/intro/CheatSheet.md
@@ -21,109 +21,109 @@ Release:
 - Is the name you assigned to an installation instance. 
 
 Revision:
-- Is the value from the Helm history command
+- Is the value from the Helm history command.
 
 Repo-name:
 - The name of a repository. 
 
 DIR:
-- Directory name/path
+- Directory name/path.
 
 ------------------------------------------------------------------------------------------------------------------------------------------------
 
 ### Chart Management
 
 ```bash
-helm create <name>                      # Creates a chart directory along with the common files and directories used in a chart.
-helm package <chart-path>               # Packages a chart into a versioned chart archive file.
-helm lint <chart>                       # Run tests to examine a chart and identify possible issues:
-helm show all <chart>                   # Inspect a chart and list its contents:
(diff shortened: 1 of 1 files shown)


scottrigby:
As part of the docs update i'm thinking we may want to remove the cheatsheet altogether (part of the convo from https://github.com/helm/helm-www/pull/1875#discussion_r2495628189). The cheatsheet leaves a lot out, and the page is not easy to keep up to date as the command flags and output changes.

One thing I do like about the Cheatsheet is the idea of categorizing different kinds of commands. If anything, it may be worth qualitatively describing these categories but linking to the commands where they live in the docs rather than copying the commands again in-place.

If we think categories of commands is useful, it may be worth discussing whether it's valuable enough to do this with the helm command and docs output itself (whether with an optional flag, etc).

Or maybe we just remove the cheatsheet in favor of the actual helm commands docs sections. Does anyone have thoughts on the value of this page?

dkarczmarski:
@scottrigby 
of course, I can do as you want,

however this is my very personal opinion:
1. I read a lot of different documentation. I have very weak memory I forget things quickly.
despite my weak memory I always remember the overall concept. so I only need to look at some example to recall it.
so that why such cheatsheet works for me well.
2. you need one hour to make a good cheatsheet. however you need one month to a 'perfect' cheatsheet. it might be not worth it.
3. when you have a cheatsheet you only need a one glance to find what you forget.
4. a cheatsheet is not a guide. it is not intended to teach you. it is only meant to recall things.

so, to sum up:
- even is it not perfect, it is still enough.
- instead of remove it, I will improve it.
- I wouldn't try to make a perfect cheatsheet.

... or maybe it would be beneficial to divide it into small pieces and include them as a part of specific topic/chapters :)

dkarczmarski:
I'm closing this PR in favour of the documentation in version 4 : #1965

PR #1868, opened 2025-11-03
fix style in intro/install.md

dkarczmarski:
correct `helm`  to `Helm` - for consistency with the rest of the document


versioned_docs/version-3/intro/install.md
```diff
@@ -111,8 +111,8 @@ sudo apt-get install helm
 ```
 
 ### From dnf/yum (fedora)
-Since Fedora 35, helm is available on the official repository.
-You can install helm with invoking:
+Since Fedora 35, Helm is available on the official repository.
+You can install Helm with invoking:
```
TerryHowe: Since you are there...
```suggestion
You can install Helm by invoking:
```

PR #1871, opened 2025-11-03
Docs: remove dollar sign to make instructions easier to copy and paste

sarafonseca-123:
Remove dollar sign to make instructions easier to copy and paste. Other sections on the document already had the $ removed, like `From Apt`.

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -18,4 +18,4 @@ jobs:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4.2.2
       - name: Check for typos
-        uses: crate-ci/typos@v1.28.3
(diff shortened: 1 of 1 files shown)


scottrigby:
I'm a little hesitant to make the v2 docs easier to copy/paste, especially given some things won't work anymore. For example, this would need to be updated to work:

```console
cd $GOPATH
mkdir -p src/k8s.io
cd src/k8s.io
git clone https://github.com/helm/helm.git
cd helm
make bootstrap build
```

and I know there's a lot more in the helm 2 instructions that no longer work. Do we really want to go down this route to make the v2 docs fully usable?

scottrigby:
also I meant to say thanks @sarafonseca-123 for the PR and appreciate contributions! There are some other areas of the docs we could really use some community help with if you'd like to dig in 🙂

TerryHowe:
I'm going to close this, sounds like it isn't happening.

PR #1872, opened 2025-11-04
Add docs page for SDK features

matheuscscp:
Docs for https://github.com/helm/helm/pull/31435.

Looks like this:

Screenshot from 2025-11-04 16-49-58

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,7 @@ See the [Overview](/overview.md) for an actionable summary of these changes.
 New features in Helm 4 that were not backported to v3
 
 ```text
+#31435  2025-11-03T18:56:14Z  matheuscscp       Introduce a context for canceling wait operations
 #31338  2025-10-21T19:51:31Z  yzewei            Add loongarch64 support
 #31351  2025-10-21T19:32:40Z  gjenkins8         feat: `helm version` print Kubernetes (client-go) version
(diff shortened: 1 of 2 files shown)


scottrigby (changes requested):
Hey @matheuscscp — thanks for jumping on this and adding docs for this SDK change 🙏 (especially since I asked 😅).

We’re currently in the middle of restructuring the Helm 4 docs and getting the bigger changes outlined. I want to make sure this lands in the right place once that framework is set, rather than having it floating on its own.

I’m going to add a hold so we don’t merge it prematurely. No action needed from you right now — we’ll pick this back up once the core layout is in and can slot it in properly.

Really appreciate you writing this up so fast — this is exactly the direction we want for the Helm 4 docs. Stay tuned!

scottrigby:
@matheuscscp we have the rest of the files in place for the v4 docs, as a work in progress. I'm thinking the best place for this content now would be in the v4 SDK section (https://helm.sh/docs/next/sdk/). That entire section needs to be not only updated but I think additional content for SDK users. Maybe that content could largely be contributed by SDK users, like you've done here. What do you think?

matheuscscp:
Sure, I can keep an eye out for relevant content as I navigate through implementing Helm 4 in Flux. Maybe @hiddeco could have some inputs here as well since he did the current integration.

matheuscscp:
> This pull request has been marked as stale because it has been open for 90 days with no activity. This pull request will be automatically closed in 30 days if no further activity occurs.

I'm going to pick this up at some point

PR #1876, opened 2025-11-05
fix: broken anchors

TerryHowe:


i18n/fr/docusaurus-plugin-content-docs/version-3/intro/install.md
```diff
@@ -20,7 +20,7 @@ et installé.
 2. Décompressez l'archive (`tar -zxvf helm-v3.0.0-linux-amd64.tar.gz`)
 3. Trouvez le binaire `helm` dans le répertoire décompréssé, et déplacez le dans la destination voulue (`mv linux-amd64/helm /usr/local/bin/helm`)
 
-A ce stade, vous devriez être cappable de lancer le client et [d'ajouter le dépot stable](/intro/quickstart.md#initialize-a-helm-chart-repository): `helm help`.
+A ce stade, vous devriez être cappable de lancer le client et [d'ajouter le dépot stable](/intro/quickstart.md#initialiser-un-dépot-de-charts-helm): `helm help`.
```
paigecalvert: Thanks for going through these @TerryHowe 

CNCF techdocs recommends a slightly different best practice for fixing broken anchor links when there's several locales that I wanted to share:

Rather than translating the anchors in the links themselves to match the given locale, the best practice is to update the heading(s) in the translated page(s) to assign English language anchor IDs.

So for example, for any broken anchor links that point to `## Initialisieren eines Helm Chart Repository`  in the german `/de/intro.quick-start.md` page, the fix would be explicitly define an anchor ID on that heading that matches the english version, like this: `## Initialisieren eines Helm Chart Repository {initialize-a-helm-chart-repository}` . This way, the links themselves don't need to be edited on a case-by-case basis

There a short explanation of this here (https://github.com/helm/helm-www/blob/main/ARCHITECTURAL_DECISIONS.md#anchor-links-to-headings) which was based on how OpenTelemetry does it (https://opentelemetry.io/docs/contributing/localization/#headings). But this reminds me that I need to update the site's page on how to localize content so this is more discoverable.

scottrigby: I added an issue to track this https://github.com/helm/helm-www/issues/1892

scottrigby (changes requested):
@TerryHowe do you want to update this PR to use this strategy https://github.com/helm/helm-www/pull/1876#discussion_r2499609355 or would you rather close this?

i18n/fr/docusaurus-plugin-content-docs/version-3/intro/install.md
TerryHowe: Fixed

PR #1881, opened 2025-11-05
fix: path to generated docs

TerryHowe:


README.md
```diff
@@ -43,7 +43,7 @@ The documentation for the list of Helm CLI Commands are [exported](https://githu
 To update these docs, you'll need to:
 
 1. Delete all plugins currently installed with `helm plugin uninstall`
-2. Navigate to `content/en/docs/helm/`
+2. Navigate to versioned_docs/version-${HELM_MAJOR_VERSION}/helm/`
```
paigecalvert: good catch @TerryHowe thanks

To update the CLI reference docs for v4, I think the command would need to be run from `docs/helm`, if I understand this process correctly.

So maybe this would say if you want to update the latest version (v4) of the CLI reference docs, navigate the `docs/helm`, and if you want to update an earlier version, go to the `versioned_docs/version-${HELM_MAJOR_VERSION}/helm/` for that version?

TerryHowe:
```suggestion
2. Navigate `docs/helm` to update the latest version or navigate to versioned_docs/version-${HELM_MAJOR_VERSION}/helm/` to update an earlier version.
```

scottrigby:
ah, i totally reworked the instructions, as there is automation to convert the helm docs to fix links, not have duplicated titles (like they did in the hugo site, etc) https://github.com/helm/helm-www/pull/1890

TerryHowe:
Take this #1890

PR #1882, opened 2025-11-05
fix: throw error on blog truncate

TerryHowe:


scottrigby:
~~I was thinking we should always add a truncate marker. I added for all old blog posts, with just enough of an initial teaser to be able to scan them in a big list (the old blog was super hard to scan or even scroll back to older entries)~~ 

edit, i misread this PR because I thought I already had it set to throw. Thanks!

PR #1895, opened 2025-11-09
Revert "Format template example"

scottrigby:
Very much appreciate contributions, but this one will bite users quickly.

Reverts helm/helm-www#1696

### DIFF
--- a/docs/howto/charts_tips_and_tricks.md
+++ b/docs/howto/charts_tips_and_tricks.md
@@ -169,9 +169,9 @@ imageCredentials:
 We then define our helper template as follows:
 ```
 {{- define "imagePullSecret" }}
-  {{- with .Values.imageCredentials }}
-    {{- printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"%s\",\"auth\":\"%s\"}}}" .registry .username .password .email (printf "%s:%s" .username .password | b64enc) | b64enc }}
-  {{- end }}
+{{- with .Values.imageCredentials }}
+{{- printf "{\"auths\":{\"%s\":{\"username\":\"%s\",\"password\":\"%s\",\"email\":\"%s\",\"auth\":\"%s\"}}}" .registry .username .password .email (printf "%s:%s" .username .password | b64enc) | b64enc }}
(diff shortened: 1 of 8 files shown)


benoittgt:
~~What is the issue ?~~

Nevermind : https://github.com/helm/helm-www/pull/1696#issuecomment-3508939309

dnskr:
As the author of the original PR, I believe Helm should encourage using indentation in templates, as the lack of it lead to one of the following real-life examples:
```yaml
{{- if and .Values.statsd.enabled }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-statsd
  labels:
    tier: airflow
    component: config
    release: {{ .Release.Name }}
    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
    heritage: {{ .Release.Service }}
{{- with .Values.labels }}
{{ toYaml . | indent 4 }}
{{- end }}
data:
  mappings.yml: |-
{{- if .Values.statsd.overrideMappings }}
    mappings:
{{ toYaml .Values.statsd.overrideMappings | indent 6 }}
{{- else }}
{{ .Files.Get "files/statsd-mappings.yml" | indent 4 }}
{{- if .Values.statsd.extraMappings }}
{{ toYaml .Values.statsd.extraMappings | indent 6 }}
{{- end }}
{{- end }}
{{- end }}
```

TerryHowe:
DCO

scottrigby:
@dnskr in actual templates yes gotemplate statements should be indented, however they should use `nindent` instead of `indent`. The issue with your initial example that this PR reverts is that it was for helper functions, not templates. Helper functions should not indent gotemplate statements. Templates should.

TerryHowe (approved):
/lgtm

dnskr:
Thank you for the explanation. I see your point, and I think it makes sense.

However, could you please provide other common examples where indentation in `.tpl` files cannot be used?
From what I’ve seen, common labels are usually the only case in a Helm chart where this applies, so I consider them an exception. I think it should be acceptable to use both approaches in one `.tpl` file, prioritizing indentation, or to separate such cases into a dedicated `.tpl` file.

PR #1899, opened 2025-11-12
New Plugins section in Helm 4 docs

scottrigby:
Initial stab at a new section for Helm 4 plugins.

Wrote this branch both to document the new Helm 4 plugin system, and also to use for the [Helm 4 Hands-on ContribFest Workshop](https://sched.co/27Nl0) at KubeCon Atlanta later today.

This PR:
- Adds a new `<GetVersion />` react component, allowing us to dynamically refer to the latest release of a specified MAJOR helm version that's configiured for the website. Get the semver from docusaurus versions configuration, optionally links to the GitHub release URL for that version.
- Adds tutorials for writing each of the 3 plugin types, using subprocess runtime. So far only the CLI plugin type tutorial also has Wasm runtime steps (the goal is to add this for the other two as well).
- Structures these under a new Helm 4 `plugins` section, with persona subsections `developer` and `user`. File name prefixes follow a subset of [Diátaxis](https://diataxis.fr/) docs system kinds, specifically `tutorial-` for the tutorials added in this PR (later could use prefixes like `how-to-`, `reference-`, etc. And we can use this ).

We still need to finish porting the relevant parts of the Helm 3 single-page Plugins doc to Helm 4, but this is something people can help with even during the ContribFest session if anyone there is motivated.

### DIFF
--- a/docs/plugins/developer/index.mdx
+++ b/docs/plugins/developer/index.mdx
@@ -6,7 +6,7 @@ sidebar_position: 3
 
 import DocCardList from "@theme/DocCardList";
 
-For an overview of Helm Plugin concepts, and how to structure and confifure plugins, read the [Plugins Overview](/plugins/overview.md).
+For an overview of Helm Plugin concepts, and how to structure and configure plugins, read the [Plugins Overview](/plugins/overview.md).
 
 This section focuses on developing Helm Plugins – it is intended to include [Tutorials](#tutorials), how-to guides, reference guides, and additional developer-focused information for writing Helm Plugins.
 
--- a/docs/plugins/overview.md
+++ b/docs/plugins/overview.md
@@ -14,7 +14,7 @@ Helm currently has 3 types of Plugins:
 
 - [CLI plugins](#cli-plugins): allow users to add additional `helm` CLI sub-commands
 - [Getter plugins](#getter-plugins): allow users to use Charts and even other Plugins in locations Helm core doesn't have built-in support for
-- [Postrenderer plugins](#postrenderer-plugins): allow users to modify Chart rendered manifests before being sent to the Kuberentes API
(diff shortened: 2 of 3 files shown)


docs/plugins/developer/index.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Plugins Developer Guide
+sidebar_label: Developing Plugins
+sidebar_position: 3
+---
+
+import DocCardList from "@theme/DocCardList";
+
+For an overview of Helm Plugin concepts, and how to structure and configure plugins, read the [Plugins Overview](/plugins/overview.md).
```
TerryHowe, line 9: General question here. I assume this document would migrate to the versioned docs and would this absolute path need to be changed?

paigecalvert: @TerryHowe for the initial release of v4, all the v4 docs could stay where they are in the main `docs` folder.

The only needed change would be marking them as the `latest` version in the docusaurus config. This would remove the "pre-release" banner from the v4 docs and also remove `/next/` from its URL path. See https://github.com/helm/helm-www?tab=readme-ov-file#how-to-move-pre-release-docs-to-ga in the readme

Then v4 would not need to be moved to the versioned_docs folder until it's time to publish the v5 pre-release docs

scottrigby: yep, done! https://github.com/helm/helm-www/pull/1902

TerryHowe: My question was really a absolute vs relative path. If this file moved to versioned docs directory, the links would need to be relative I think.

scottrigby: For now, all markdown links need to be absolute paths. The syntax changes depending on whether the link is within or between docs and blog. Paige updated the ARCHITECTURAL_DECISIONS.md doc with details about this: https://github.com/helm/helm-www/blob/main/ARCHITECTURAL_DECISIONS.md#markdown-links

PR #1907, opened 2025-11-12
Replace get-helm-3 with get-helm-4

richardmcsong:


docs/intro/install.mdx
```diff
@@ -4,10 +4,6 @@ description: Learn how to install and get running with Helm.
 sidebar_position: 2
 ---
 
-import Helm4 from "../_v4-in-progress.mdx" 
-
-<Helm4/>
```
scottrigby, line 9: before removing this line there is one more mention under "from source" about the "latest (pre-release)" of Helm. you'll want to remove the pre-release parentheses now.

scottrigby (changes requested):

PR #1916, opened 2025-11-13
Update SDK examples for Helm v4

gjenkins8:


i18n/es/docusaurus-plugin-content-docs/version-3/topics/advanced.md
```diff
@@ -70,8 +70,8 @@ import (
     "log"
     "os"
 
-    "helm.sh/helm/v3/pkg/action"
-    "helm.sh/helm/v3/pkg/cli"
+    "helm.sh/helm/v4/pkg/action"
+    "helm.sh/helm/v4/pkg/cli"
```
scottrigby: we don't want to change v3 -> v4 for any of these translation pages under `version-3`, right?

gjenkins8: Bad search/replace (I excluded `versioned_docs/` dir, but not these translation dirs). Will revert

sdkexamples/install.go
```diff
@@ -21,14 +22,14 @@ func runInstall(ctx context.Context, logger *log.Logger, settings *cli.EnvSettin
 
 	installClient := action.NewInstall(actionConfig)
 
-	installClient.DryRunOption = "none"
+	installClient.DryRunStrategy = "none"
+	installClient.WaitStrategy = "watcher"
```
benoittgt, line 26: Don't you need non-zero wait timeout for this?

This is something that could be [ommited](https://github.com/helm/helm/pull/31421/files#diff-a77e457a3f361ea0907b9f8898d9c1aa33a3e9286038821969a3ea008f49faaeR55) with https://github.com/helm/helm/pull/31421

scottrigby: @benoittgt with this PR, sdk examples run fine without setting timeout. Approving this based on my understanding, but clarify if you meant something else?

That said, https://github.com/helm/helm/pull/31421 is now merged with a `needs docs` label, and should be part of helm `4.1.0`. A follow-up PR could be made to make use of that new feature - and note it somewhere in the sdk dosc section / sdk examples, and add the `awaiting release` label, noting that we should wait until 4.1.0 is released merge on that one. Would you want to add this follow-up PR?

gjenkins8: Should add the timeout for completeness. Will make another PR (get this one merged)

Issue #1913, opened 2025-11-13
RSS feed broken


stevehipwell:
The website RSS feed is currently broken, this happened before 2025-11-04 as the last post isn't in my feed reader.

TerryHowe:
11/4 was the last post in the feed https://helm.sh/blog/rss.xml

Was it reporting differently before?

stevehipwell:
@TerryHowe it looks like the URL changed from https://helm.sh/blog/index.xml to https://helm.sh/blog/rss.xml between the last two posts which broke my feed reader.

stevehipwell:
@TerryHowe the new feed doesn't look correct as historic posts are showing up as 2 weeks old and all posts are redirecting to localhost:3000/

TerryHowe:
Thanks for getting those details. It probably wouldn't be hard to at least change the URL with docusaurus configuration. Do you want to raise a PR? I'm headed out for vacation, so I'll be a bit limited for the next several days.

stevehipwell:
https://github.com/helm/helm-www/blob/6e6fb02783ee8d378405e3ee935b5347d48c217a/docusaurus.config.js#L9

It looks to me like the `SITE_URL` env variable hasn't been set?

Andygol:
IMO the issue is in the hardcoded vars in the Makefile 🤷 

https://github.com/helm/helm-www/blob/f257c3b60124eb4d4198d18bb293ba5841ddef1e/Makefile#L1-L2

This must help 👇 

```diff
-SITE_URL := http://localhost:3000
-BASE_URL := /
+SITE_URL ?= http://localhost:3000
+BASE_URL ?= /
```

TerryHowe:
@Andygol the RSS feed had the correct URL, the path just changed in the Docusaurus migration. The PR https://github.com/helm/helm-www/pull/1948 redirects the old path to the new.

PR #1948, opened 2025-11-20
fix: redirect old rss feed to new

TerryHowe:
https://deploy-preview-1948--helm-merge.netlify.app/blog/index.xml
https://deploy-preview-1948--helm-merge.netlify.app/blog/rss.xml

Closes: https://github.com/helm/helm-www/issues/1913

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -50,6 +50,12 @@ package = "./netlify-plugins/cache-docusaurus-dirs-file"
   to = "https://docs.google.com/document/d/10cyzggPiLeLz169cDF1UaUMmM4uZx6rO3Bw9rRQRjJA/edit?usp=sharing"
   status = 302
 
+# RSS feed alias
+[[redirects]]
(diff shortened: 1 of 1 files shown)


scottrigby:
Question, did the blog rss feed previously work for languages? For example, this works on the current site https://deploy-preview-1948--helm-merge.netlify.app/ko/blog/rss.xml, but this doesn't https://deploy-preview-1948--helm-merge.netlify.app/ko/blog/index.xml. I'm not sure if it ever did though.

TerryHowe:
> Question, did the blog rss feed previously work for languages? For example, this works on the current site https://deploy-preview-1948--helm-merge.netlify.app/ko/blog/rss.xml, but this doesn't https://deploy-preview-1948--helm-merge.netlify.app/ko/blog/index.xml. I'm not sure if it ever did though.


No, the previous site did not i18n the blog. Personally, I don't think we should i18n the blog just because of value/complexity.

PR #1951, opened 2025-11-24
Update version for get-helm link

ayacoo:

PR #1952, opened 2025-11-24
fix: installing faq reference to get-helm-4

TerryHowe:


scottrigby:
ref https://github.com/helm/helm/pull/31485

TerryHowe:
I'm just trying to fix todays problem here although I agree this FAQs is pretty useless.

paigecalvert:
> I'm thinking we should move these (now very minimal) FAQs to the "Installing Helm" page https://helm.sh/docs/intro/install, and do away with the FAQ section altogether. This Script is already updated on that page as well by #1951.
> 
> WDYT?
> 
> CC @paigecalvert since I know you're working on a content reorganization proposal?

Agree. (Going to get a draft pr open in the cncf repo for that docs analysis today. Finished most of the eval last week and just making it through the recommendations portion of the exercise 😄 )
  
I also have no problem with fixing this FAQ as is for now.

TerryHowe:
Created https://github.com/helm/helm-www/issues/1966 to track that

PR #1953, opened 2025-11-24
chore: update-docs.sh script run on v3

TerryHowe:
This is what I get when I run `update-docs.sh` which is in https://github.com/helm/helm-www/pull/1767


TerryHowe:
This must have been done.

PR #1960, opened 2025-11-27
Return atomic flag in the upgrade docs

mynoheart:
I've noticed that the third version of upgrade command contains a new flag. But the previous one was deleted.

### DIFF
--- a/docs/chart_template_guide/function_list.mdx
+++ b/docs/chart_template_guide/function_list.mdx
@@ -1457,7 +1457,7 @@ dictionary:
 Given:
 
 ```
-dst:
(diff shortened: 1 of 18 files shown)


mynoheart:
Hmm... I thought that it could be generated but I expected a different folder for that. Like `dist` or something. :-) Let's check later a little bit.

mynoheart:
And the button drives to edit the files in place. :-)

image

mynoheart:
I regenerated CLI docs and updated the current versions.

TerryHowe:
Done in another PR :-(

PR #1975, opened 2025-12-31
Use --dry-run=client in values_files.mdx

doncicuto:
According to helm warnings, `--dry-run is deprecated and should be replaced with '--dry-run=client'`, so I propose to add the client value to both dry-run commands examples

Thanks!


TerryHowe (approved):
/lgtm

PR #1978, opened 2026-01-03
Fix KubeCon NA event showing as upcoming

### LINKED ISSUE #1977: [Docs]: Move KubeCon NA 2025 to past events
**Scope:** 

Was surfing through the landing page at https://helm.sh, found that KubeCon NA still shows as 'upcoming'. 

**Intended Behavior:**

Should be moved into 'Past Events'.

devdrx:
Fixes #1977  

### What
Moves KubeCon NA 2025 from Upcoming to Past Events on the Helm website.

### Why
KubeCon NA 2025 has already occurred but was still shown as upcoming.

### How
Updated the upcoming event listing with 2026 EU event and pushed the 2025 NA event into the past events list.

## Before:
image

## After:
image

### DIFF
--- a/src/components/HomeCommunity/index.js
+++ b/src/components/HomeCommunity/index.js
@@ -114,9 +114,9 @@ const BlockList = [
             </Translate>
           </dt>
           <dd>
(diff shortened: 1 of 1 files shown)


TerryHowe (commented):
/lgtm

PR #1979, opened 2026-01-06
Update install.mdx

s7santosh:
Remove the $ prompt character from the command


TerryHowe (changes requested):
For consistency, please update the "### From Source (Linux, macOS)" section in this file.

There has been resistance to some similar changes in the past, but I can't imagine right now that would be.

Issue #1752, opened 2025-08-19
Quickstart Guide References Bitnami Charts, is this affected by "Upcoming changes to the Bitnami catalog (effective August 28th, 2025)"


nelsonjchen:
https://helm.sh/docs/intro/quickstart/

https://github.com/bitnami/charts/issues/35164

Admittedly, I'm still a bit green to some of this, but I'm unclear if this will affect things. Looking at the [`bitnami/mysql` chart](https://artifacthub.io/packages/helm/bitnami/mysql), I think it references a non-`:latest` tag. Is that going to break come August 28th?

benoittgt:
I think it still need to be look at.

TerryHowe:
Anyone willing to work on this? I'm not a bitnami user ATM, so I don't know all the impacts.

Iam-Karan-Suresh:
Hi @TerryHowe ! I’d like to take a look at this issue and work on it, if that’s okay.

From my understanding, the main concern is that the Helm Quickstart
currently depends on Bitnami charts, which may become confusing or fragile
due to upcoming Bitnami catalog and image changes (e.g. legacy image overrides).

My proposed approach would be:
- Avoid using Bitnami charts in the main Quickstart flow
- Update the example to use a vendor-neutral, OCI-based chart (for example via GHCR)
- Add a short note acknowledging Bitnami and other third-party providers, so existing users aren’t confused

This keeps the Quickstart stable and beginner-friendly while still pointing users
to third-party options intentionally.

If this approach sounds reasonable, I’m happy to open a PR.
Please let me know and feel free to assign the issue to me.
Thanks!

lambdasierra:
I can confirm that the current quick-start instructions did not work for me.

Adding the repository works:

https://github.com/helm/helm-www/blob/17ce85a09ccc1f198fa622d3fc81cc51e7f46615/docs/intro/quickstart.md#L41

The repository update works:

https://github.com/helm/helm-www/blob/17ce85a09ccc1f198fa622d3fc81cc51e7f46615/docs/intro/quickstart.md#L63

The installation *appears* to work:

https://github.com/helm/helm-www/blob/17ce85a09ccc1f198fa622d3fc81cc51e7f46615/docs/intro/quickstart.md#L64

```shell
$ helm install bitnami/mysql --generate-name
NAME: mysql-1767987296
```

although it prints the following warning:

```
⚠ WARNING: Since August 28th, 2025, only a limited subset of images/charts are available for free.
[...]

PR #1982, opened 2026-01-13
docs: make Quickstart vendor-neutral using helm create

Iam-Karan-Suresh:
### Summary

This PR updates the Helm Quickstart Guide to avoid relying on Bitnami charts
and classic Helm repositories in the main onboarding flow.

The guide now demonstrates Helm fundamentals using a locally generated chart
via `helm create`, while still acknowledging third-party repositories and
OCI-based charts as valid options.

### Motivation

Bitnami has announced upcoming catalog and image changes, which may require
additional configuration (such as legacy image overrides) and can be confusing
for new users following the Quickstart.

Since the Quickstart is intended to be a stable, beginner-friendly introduction
to Helm, depending on a specific third-party vendor introduces unnecessary
fragility.

### What changed

- Removed Bitnami-specific commands from the main Quickstart flow
- Replaced the example with a vendor-neutral `helm create` → `helm install` workflow
- Added a short note acknowledging Bitnami and other third-party repositories
- Clarified that Helm supports both local charts and OCI-based registries

### Why `helm create`

Using `helm create`:
- Avoids external dependencies
- Works consistently in all environments
- Teaches core Helm concepts (charts, installs, releases)
- Aligns with the suggestion raised in the original issue discussion

### Scope

This is a documentation-only change and does not modify Helm behavior.

Feedback and suggestions are welcome.
issue: #1752

### DIFF
--- a/docs/intro/quickstart.md
+++ b/docs/intro/quickstart.md
@@ -31,51 +31,42 @@ or look at [the official releases page](https://github.com/helm/helm/releases).
 
 For more details, or for other options, see [the installation guide](/intro/install.mdx).
 
-## Initialize a Helm Chart Repository
+## Install an Example Chart
 
-Once you have Helm ready, you can add a chart repository. Check [Artifact
+Once you have Helm ready, you typically add a chart repository. Check [Artifact
 Hub](https://artifacthub.io/packages/search?kind=0) for available Helm chart
 repositories.
 
-```console
-$ helm repo add bitnami https://charts.bitnami.com/bitnami
-```
-
-Once this is installed, you will be able to list the charts you can install:
-
-```console
-$ helm search repo bitnami
-NAME                             	CHART VERSION	APP VERSION  	DESCRIPTION
-bitnami/bitnami-common           	0.0.9        	0.0.9        	DEPRECATED Chart with custom templates used in ...
-bitnami/airflow                  	8.0.2        	2.0.0        	Apache Airflow is a platform to programmaticall...
-bitnami/apache                   	8.2.3        	2.4.46       	Chart for Apache HTTP Server
-bitnami/aspnet-core              	1.2.3        	3.1.9        	ASP.NET Core is an open-source framework create...
-# ... and many more
-```
-
-## Install an Example Chart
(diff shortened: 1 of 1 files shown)


docs/intro/quickstart.md
```diff
@@ -33,49 +33,42 @@ For more details, or for other options, see [the installation guide](/intro/inst
...
 Hub](https://artifacthub.io/packages/search?kind=0) for available Helm chart
 repositories.
 
-```console
-$ helm repo add bitnami https://charts.bitnami.com/bitnami
-```
+> **Note:** Popular repositories include Bitnami and other third-party providers. While these are excellent sources for production-ready charts, this guide will focus on creating a local chart or using an OCI-based chart to ensure a stable, self-contained quickstart experience.
 
-Once this is installed, you will be able to list the charts you can install:
+## Install an Example Chart
```
TerryHowe:
```suggestion
```

docs/intro/quickstart.md
```diff
@@ -33,49 +33,42 @@ For more details, or for other options, see [the installation guide](/intro/inst
 
 ## Initialize a Helm Chart Repository
```
TerryHowe:
```suggestion
## Install an Example Chart

```

Iam-Karan-Suresh:
Hi @TerryHowe,
I have updated the PR according to your suggestions. Kindly review when you have time. Thanks!

TerryHowe (approved):
/lgtm

gjenkins8 (changes requested):
While I completely agree with avoiding e.g. bitnami as an example chart source (times have certainly changed a lot since bitnami was a good example source). I do think the quick start should be a more realistic basic user workflow - that should involve installing an external chart.

The majority of Helm users *are not* chart produces, and using `helm create ...` here is both a distraction and doesn't show the user how to use an external chart.

For a "vendor neutral" source today, how about https://artifacthub.io/packages/helm/podinfo/podinfo ? We should certainly promote an OCI source, then a follow up section on using a legacy repository (perhaps).

TerryHowe:
Dup

PR #1993, opened 2026-01-27
Removing regular user prompt symbol

JoshuaCooper:
Copy/Paste from this doc brings $ results in the following:

```
$: command not found
$: command not found
$: command not found
```


TerryHowe:
I merged the other outstanding PR on this topic, so this was a dup. Thanks though!

PR #2001, opened 2026-02-12
Adds a writing style guide for the helm docs

paigecalvert:
Per [HIP-0024](https://helm.sh/community/hips/hip-0024), this PR adds an initial writing style guide for the Helm documentation. A style guide is an important step for helping the community to be able to write consistent, clear documentation.

This PR proposes that the Helm documentation defaults to the [Google developer documentation style guide](https://developers.google.com/style), which is an industry-standard style guide and often referenced by the CNCF techdocs team. It also adds a new `style-guide.md` page to provide a summary of key guidelines from Google's style guide, plus some Helm-specific guidelines not covered in Google's style guide.

These are intended to be guidelines, not rules. Additionally, the aim is that the Helm community continue to iterate on these style guidelines as needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,7 +202,8 @@ Like any good open source project, we use Pull Requests (PRs) to track code chan
 
 Documentation PRs will follow the same lifecycle as other PRs. They will also be labeled with the
 `docs` label. For documentation, special attention will be paid to spelling, grammar, and clarity
-(whereas those things don't matter *as* much for comments in code). For additional style guidelines, see [Documentation Style Guide](style-guide.md).
+(whereas those things don't matter *as* much for comments in code).
+For additional style guidelines, see [Documentation Style Guide](style-guide.md).
 
 ## The Triager
 
--- a/style-guide.md
+++ b/style-guide.md
@@ -2,7 +2,7 @@
 
 This page lists writing style guidelines for the Helm documentation.
 These are guidelines, not rules.
-Use your best judgement, and depart from these guidelines when doing so improves your content.
+Use your best judgement, and depart from these guidelines when doing so improves our content.
 
 By default, the Helm docs use the Google Developer Docs Style Guide: https://developers.google.com/style/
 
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -202,7 +202,7 @@ Like any good open source project, we use Pull Requests (PRs) to track code chan
 
 Documentation PRs will follow the same lifecycle as other PRs. They will also be labeled with the
 `docs` label. For documentation, special attention will be paid to spelling, grammar, and clarity
-(whereas those things don't matter *as* much for comments in code).
+(whereas those things don't matter *as* much for comments in code). For additional style guidelines, see [Documentation Style Guide](style-guide.md).
```
scottrigby: Can we use [Semantic Line Breaks](https://sembr.org/) when adding this new sentence to the paragraph?

```suggestion
(whereas those things don't matter *as* much for comments in code).
For additional style guidelines, see [Documentation Style Guide](style-guide.md).
```

i realize we use mixed styles in our docs now.

style-guide.md
```diff
@@ -0,0 +1,51 @@
...
+- Avoid time-bound terminology like "currently", "new", "at this time", and "now". Instead, write timeless documentation that makes no assumptions about a reader's prior knowledge.
+- Avoid using the word "should". Instead, strive for more prescriptive documentation. For example, if an action is required, use words and phrase like "must" or "ensure that". If an action is optional/recommended, use words and phrases like "you can", "might", or "we recommend".
+
+## Readability
+
+- Break up walls of text to make content easier to scan
+- Try to use fewer than 26 words per sentence
+- Define acronyms and abbreviations on first usage
+- Procedural/how-to content must use numbered steps. For one-step procedures, use a bullet point. See https://developers.google.com/style/procedures#single-step-procedures for examples
+
```
scottrigby, line 52: should we add a new bullet here about semantic line breaks for markdown, to make it easier to read raw files and git diffs?

style-guide.md
```diff
@@ -0,0 +1,51 @@
+# Documentation Style Guide
+
+This page lists writing style guidelines for the Helm documentation.
+These are guidelines, not rules.
+Use your best judgement, and depart from these guidelines when doing so improves your content.
```
TerryHowe:
```suggestion
Use your best judgement, and depart from these guidelines when doing so improves our content.
```

CONTRIBUTING.md
TerryHowe: Should semantic line breaks be mentioned? It generally make review easier.

TerryHowe (commented):
Looks good !

CONTRIBUTING.md
paigecalvert: Yes good call. I'll add semantic line breaks to the style guide

paigecalvert:
> i like this! some inline comments, and a few more thoughts here:
> 
> should we also point to docusaurus docs on syntax it supports on top of common mark (admonitions, codeblock additional options, etc)?

good idea! forgot to add this 

> should we also note style guidelines for reusable content, and embedded files using mdx file extension? 
> or is this doc generally about the guidelines, and other files are more about the “how”?

That's a good question...I think your point about guidelines versus "the how" might be right. Could be good to consider for the contributing guide?

paigecalvert:
@scottrigby @TerryHowe thank you for your reviews! I addressed your comments to add info about the markdown features supported by docusaurus and semantic line breaks.

I also added one more guideline about using notes and asides, which I was reminded of after talking about docusaurus admonitions (specifically "Use notes and asides sparingly. In general, the user should be able to succeed even if they skip the information in the note.").

PR #2056, opened 2026-03-30
Add troubleshooting entry for rollback missing resource error

promptless-for-oss:
Documents the scenario where `helm rollback` fails with "no [Resource] with the name '[name]' found" when a failed upgrade leaves resources on the cluster that are missing from the original release metadata. Explains the specific sequence of events that triggers this issue and notes that it is fixed in newer Helm 3 releases.


TerryHowe (approved):
/lgtm

promptless-for-oss:
Just a reminder: If you'd like me to act on any feedback you have via Github comments, just type @promptless in your suggestion and I'll get right on it! (I won't show up in the user dropdown, but I'll process any request that has @promptless in the comment body.)

PR #2057, opened 2026-04-01
Clarify helm registry login host format requirements

promptless-for-oss:
Documents that the host parameter for `helm registry login` must be only the registry hostname with an optional port, without URL schemes (http://, https://) or paths. Adds a table with valid and invalid examples to the OCI registries documentation.

### DIFF
--- a/docs/topics/registries.mdx
+++ b/docs/topics/registries.mdx
@@ -55,6 +55,15 @@ Password:
 Login succeeded
 ```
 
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

promptless-for-oss:
Just a reminder: If you'd like me to act on any feedback you have via Github comments, just type @promptless in your suggestion and I'll get right on it! (I won't show up in the user dropdown, but I'll process any request that has @promptless in the comment body.)

PR #2065, opened 2026-04-10
Add troubleshooting entry for helm pull --debug HTTP output

promptless-for-oss:
Documents the --debug flag for helm pull, which now outputs HTTP-level debug information (URL being fetched and response status) to help users troubleshoot connectivity, authentication, or repository URL issues.


TerryHowe:
Wrong directory and not worth the squeeze.

PR #2082, opened 2026-04-20
Clarify with scope note in control structures docs

### LINKED ISSUE #2070: statement in documentation
On this page:

https://helm.sh/docs/chart_template_guide/control_structures/

It says

"Note that we removed the if conditional from the previous exercise because it is now unnecessary"

The if conditional in the previous exercise discussed "coffee" and "mug".  

With the "if conditional" removed it will not show "mug".  

"mug" is gone.   

With that in mind, how was the "if conditional" "unnecessary"?  It was necessary to display "mug" right?

karthi-palanisamy:
The previous note said the if conditional was removed because it was "unnecessary", which was confusing since the mug example relied on it. Reword to clarify that the example was simplified to focus on with scoping, and that with itself acts as an empty value guard.

Fixes #2070

### DIFF
--- a/docs/chart_template_guide/control_structures.md
+++ b/docs/chart_template_guide/control_structures.md
@@ -285,9 +285,10 @@ data:
   {{- end }}
 ```
 
-Note that we removed the `if` conditional from the previous exercise
-because it is now unnecessary - the block after `with` only executes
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

PR #2087, opened 2026-04-23
Enhance .helmignore documentation with command details

### LINKED ISSUE #1171: Misleading documentation of .helmignore
Within the documentation for the `.helmignore` file [[1]], it is is stated that:

>The .helmignore file is used to specify files you don't want to include in your helm chart.
>
> If this file exists, the **helm package** command will ignore all the files that match the pattern specified in the .helmignore file while packaging your application.

However, the .helmignore file is also interpreted if running **helm install** or **helm upgrade**.

**Expected behavior**
Files are only ignored if running **helm package** but interpreted if running other sub-commands, like **install**.

**Current behavior**
Files are always ignored independent of the sub-command.

**Use case**
I do have a sub-folder named *ca-certifcates* within my helm directory. 
The Helm chart includes a secret specification, which automatically add all files within this directory as secret-data.
This allows the user to customize the containers to handle connection to servers with self-signed certificates etc.

The secret specification is:
```
---
apiVersion: v1
kind: Secret
metadata:
  name: ca-certificates
data:
{{ (.Files.Glob "ca-certificates/*").AsSecrets | indent 2 }}
```
The secret is then mounted into the containers which in turn have an entrypoint handling those files and adding them to the CA store of the OS.

Using .helmignore, the files are ignored even for **install** and **upgrade**, which renders the secret to being empty.

**Current workaround**
Prior to running **helm package** I manually remove all files not needed within the Helm chart and leave the .helmignore empty,

**Notes**
Are there any other workarounds, which allows to have files within the directory structure of Helm, but do not include them when running the **package** sub-command?

[1]: https://helm.sh/docs/chart_template_guide/helm_ignore_file/

ringerc:
Expanded the explanation of the `.helmignore` file to clarify its impact on various Helm commands and highlighted differences from `.gitignore`. Add an explicit warning about broken negation. Link to [HIP 27](https://github.com/helm/community/blob/main/hips/hip-0027.md) to surface proposed changes.

Fixes https://github.com/helm/helm-www/issues/1171

Improves docs to clarify behaviour referred to in https://github.com/helm/helm/issues/8688, https://github.com/helm/helm/issues/3622

### DIFF
--- a/docs/chart_template_guide/helm_ignore_file.md
+++ b/docs/chart_template_guide/helm_ignore_file.md
@@ -7,16 +7,17 @@ sidebar_position: 12
 The `.helmignore` file is used to specify files you don't want to include in
 your helm chart.
 
-If this file exists, the `helm package` command will ignore all the files that
-match the pattern specified in the `.helmignore` file while packaging your
-application.
+If this file exists, Helm commands will ignore all the files that
+match the pattern specified in the `.helmignore` file. The ignore file affects
+most Helm commands including `helm package`, `helm template` and `helm install`.
 
 This can help in avoiding unnecessary or sensitive files or directories from
 being added in your helm chart.
 
 The `.helmignore` file supports Unix shell glob matching, relative path
-matching, and negation (prefixed with !). Only one pattern per line is
-considered.
+matching. Only one pattern per line is considered. 
+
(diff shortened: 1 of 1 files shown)


TerryHowe:
Reopen if you have something, there is some content here worth saving, but no activity and conflicts and typos failure

PR #2088, opened 2026-04-23
fix: use absolute paths for calendar and blog links, add htmltest config

TerryHowe:
## Summary

- Fix `href="calendar/release"` → `href="/calendar/release"` in `HomeCommunity` component — the relative path breaks on all non-English locale pages (e.g. from `/ja/`, the link resolves to `/ja/calendar/release` which has no Netlify redirect; only `/calendar/release` is redirected to Google Calendar)
- Fix `href="blog/helm-dco"` → `href="/blog/helm-dco"` — same relative path issue across all i18n locale homepages
- Add `.htmltest.yml` to configure `bin/htmltest` for this project, eliminating ~383K false-positive errors from `htmltest` while keeping meaningful checks:
  - `IgnoreDirectoryMissingTrailingSlash: true` — Docusaurus builds directory-based URLs; Netlify handles trailing-slash redirects natively, so these are not real broken links
  - `CheckInternalHash: false` — suppresses false positives from `href="#"` mobile nav toggles; Docusaurus already validates real anchor links at build time via `onBrokenLinks: "throw"`
  - `IgnoreURLs: ["^(/[a-z]+)?/404$"]` — Docusaurus builds only one global `404.html`; locale-specific 404 pages (`/de/404`, `/ja/404`, etc.) in the language switcher do not exist as files but are handled at the CDN/server level

## Test plan

- [ ] `bin/htmltest --skip-external` passes with 0 errors after rebuilding (the two link fixes require a rebuild to reflect in the built HTML)
- [ ] Verify `/calendar/release` still redirects to Google Calendar on the deployed preview
- [ ] Verify `/blog/helm-dco` page loads correctly from all locale homepages

🤖 Generated with [Claude Code](https://claude.com/claude-code)


paigecalvert (approved):
lgtm

PR #2106, opened 2026-05-14
Chart Tests: small wording modification

sdarwin:
On the Chart Tests page, the text previously said "a test is... a job definition."  
The problem with this, is that "Job" is an official kubernetes concept.      
A manifest might be "kind: Job".  If you are reading k8s documentation and see the word "job" it means something.  So this PR modifies the sentence slightly.


TerryHowe (approved):
/lgtm

yxxhero (approved):

PR #2119, opened 2026-05-27
Upgrade Docusaurus to 3.10.1

paigecalvert:
## Summary

- Upgrades Docusaurus from 3.9.2 to 3.10.1
- Adds @docusaurus/faster, which is required for 3.10.x when `future: {v4: true,},`. This enables a modernized build infrastructure for much faster builds. See https://docusaurus.io/blog/releases/3.10#docusaurus-faster
- Converts heading IDs for all MDX files to use the MDX comment syntax (`{/* #heading-id} */`). This fixes broken builds in several locales that were using the classic syntax (`{#heading-id}`) . See https://docusaurus.io/docs/markdown-features/toc#heading-ids
- Adds a script to `scripts` for converting heading IDs from the classic syntax to the MDX comment syntax.


scripts/convert-mdx-anchor-ids.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
```
paigecalvert, line 1: We may or may not this, but this is what I used to convert the syntax of explicit MDX heading anchors from `{#id}` to `{/* #id */}`.

AFAIK, using the "classic" syntax for heading IDs in MDX files did not break any builds on Docusaurus 3.9.x, but I was seeing broken builds caused by this after upgrading to 3.10.1.

Also, theoretically the [`write-heading-ids`](https://docusaurus.io/docs/cli#docusaurus-write-heading-ids-sitedir) CLI would be able to handle the syntax conversions, but I needed something that would only convert the heading ID syntax in MDX files since we have a mix of MD and MDX in the site.

TerryHowe (approved):
/lgtm

PR #2122, opened 2026-05-28
APT installation validate gpg key ID

gjenkins8:
The APT installation instructions download the repository signing GPG key from the same source as the GPG mirror itself. Should the repository be compromised, any user to initializes the repository configuration would end up trusting a compromised signing key

(and reinitialization is common, as today users will frequent initialize a repository during a Container or VM build e.g. Dockerfile/Containerfile or Ansible configuration run; compared to historically, where an "admin" might have initialized the key on a system once).

This PR includes a fragment to validate the GPG Key ID matches the (currently) expected Key ID / fingerprint during the repository initialization. And to error if not.

### DIFF
--- a/versioned_docs/version-3/intro/install.md
+++ b/versioned_docs/version-3/intro/install.md
@@ -112,10 +112,8 @@ curl -fsSL https://packages.buildkite.com/helm-linux/helm-debian/gpgkey > "${TMP
 # Ensure that the key ID matches to prevent a repository compromise from establishing an attacker controlled key
 if [ "$(gpg --show-keys --with-colons "${TMPDIR:-/tmp}/helm.gpg" | awk -F: '$1 == "fpr" {print $10}' | head -n 1)" != "${HELM_BUILDKITE_APT_KEY_ID}" ]; then echo "ERROR: Unexpected Helm APT key ID: potential key compromise"; exit 1; fi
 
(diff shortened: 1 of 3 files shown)


versioned_docs/version-3/intro/install.md
```diff
@@ -102,10 +102,19 @@ winget install Helm.Helm
 Members of the Helm community have contributed an Apt package for Debian/Ubuntu. This package is
 generally up to date. Thanks to [Buildkite](https://buildkite.com/organizations/helm-linux/packages/registries/helm-debian) for hosting the repo.
 
-```console
+```shell
```
sabre1041: Since this is for the versioned docs and Helm 3, performing these commands (many of them on this page) will result in an install of Helm 4. Is that a concern?

sabre1041 (approved):
Approving as specifically scope the validation of GPG ID

versioned_docs/version-3/intro/install.md
gjenkins8: Other way around (from a quick test). This repo still points to Helm 3. Vaguely, I remember discussion around a new v4 repo (so users have to opt-in to upgrade). Overall, a separate PR / issue to fix this I think (this PR is just adding the signature validation)

TerryHowe (approved):
/lgtm

PR #2126, opened 2026-05-30
Release checklist separate minor/patch instruction flows

gjenkins8:
Separate minor/major and patch instruction these into separate lists,  since these are practically done as independent operations. As the current release checklist interleaves the major/minor and patch release checklist items.

There is some duplication of instructions, but these are easily manageable IMHO.


banjoh (approved):

TerryHowe (approved):
/lgtm

PR #2129, opened 2026-06-04
build: enable @docusaurus/faster build optimization

### LINKED ISSUE #2120: @docusaurus/faster is installed but never enabled
## Problem

`@docusaurus/faster` is installed but never enabled.

`package.json` adds `@docusaurus/faster@^3.10.1` (see #2119), and the PR description claims this "enables a modernized build infrastructure for much faster builds." But `docusaurus.config.js` only sets `future: { v4: true }` — there is no `future: { faster: true }`.

Per the [3.10 release notes](https://docusaurus.io/blog/releases/3.10), Faster is opt-in via `future.faster: true` (the v4-compat flag does **not** auto-enable it). As written, the dependency is dead weight and the promised speedup won't happen.

## Fix

Either:
- Add `faster: true` to the `future` block in `docusaurus.config.js`, or
- Drop the `@docusaurus/faster` dependency.

somaz94:
`@docusaurus/faster@^3.10.1` was added to `package.json` (#2119) but never enabled — `docusaurus.config.js` only sets `future: { v4: true }`, so the Rspack / SWC / Lightning CSS optimizations the dependency provides are currently dead weight. Per the [Docusaurus 3.10 release notes](https://docusaurus.io/blog/releases/3.10), Faster is opt-in via `future.faster: true` and is **not** auto-enabled by the `v4` compatibility flag.

This sets `future.faster: true`, activating the build optimization the dependency was installed for.

### Validation

- `yarn build --locale en` → `[SUCCESS] Generated static files in "build". Done in 9.87s.` (exit 0)
- Confirmed `@rspack` is engaged in the build with the flag enabled (Faster active, not dead weight).
- The only build warning is a pre-existing broken anchor (`/docs/intro/install` → quickstart) unrelated to this change.

closes: #2120

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -42,6 +42,7 @@ const config = {
   // Future flags, see https://docusaurus.io/docs/api/docusaurus-config#future
   future: {
     v4: true, // Improve compatibility with the upcoming Docusaurus v4
+    faster: true, // Enable @docusaurus/faster build optimizations (Rspack, SWC, Lightning CSS)
   },
 
   // Opt-in to less strict, standard CommonMark support with options


paigecalvert:
See https://github.com/helm/helm-www/issues/2120#issuecomment-4623951308

somaz94:
Closing this — after reproducing the test myself, I confirmed the Faster optimizations are already active on `main` without this change, so the explicit `faster: true` is a no-op here. Thanks @paigecalvert for catching it.

## What I verified

I checked out `main` (which has only `future: { v4: true }`, no explicit `faster`) and ran the same perf build:

```bash
DOCUSAURUS_PERF_LOGGER=true yarn build --locale en
```

Rspack is already engaged — no `faster: true` required:

```
[PERF] Build > en > Creating rspack bundler configs - 103.85 ms
[PERF] Build > en > Bundling with rspack - 4.72 seconds!
[SUCCESS] Generated static files in "build".
```

(No webpack anywhere in the logs.)

## Why `v4: true` is sufficient

In Docusaurus 3.10, `future: { v4: true }` is a shorthand that turns on all the v4 sub-flags, one of which is `fasterByDefault`. Per the [config reference](https://docusaurus.io/docs/api/docusaurus-config#future), `fasterByDefault` *"Defaults all `future.faster` flags to true instead of false. Requires having `@docusaurus/faster` in your dependencies."*

Since `@docusaurus/faster` was already added in #2119, the chain `v4: true → fasterByDefault → future.faster.* (incl. rspackBundler)` is already satisfied. Adding `faster: true` explicitly changes nothing.

So the dependency from #2119 isn't dead weight — it's already doing its job. Closing as redundant; this also resolves the premise of #2120.

PR #2145, opened 2026-06-18
Helm 4 banner drop minor/patch versions

gjenkins8:


TerryHowe (approved):
/lgtm

Should at some point call attention to v3 eol

PR #2165, opened 2026-07-03
Document how to prevent YAML-injection in Chart Templates? [2]

meeque:
This is an alternative PR for #2154.

I had the impression that previous PR #2158 didn't fit well with existing contents of the "YAML Techniques" appendix.

This PR attempts to address this by editing that appendix more heavily. Major changes are:

* Split YAML types section into three:
  * *Strings*
  * *Other Scalars*
  * *Collections*
* Move *Collections* section to the end of these, since collections often use scalars.
  (Not sure about this decision, can revert if original order is deemed preferable.)
* For each section about YAML types, add a subsection about generating these types from dynamic values.
* Add a security considerations section at the the bottom.

I know this is a big edit, sorry for the spam. But I feel it conveys what template authors need to know about generating YAML.


meeque:
Note: I've tried to stick to the existing writing style of the appendix. I know that it contradicts the style-guide, but imho that should be addressed through a separate issue...

TerryHowe:
Dup, reopen if you have corrections/improvements to the current, thanks.

PR #2175, opened 2026-07-10
docs(i18n): add Korean Go data types guide

BusanGukbap:
## Summary

- Add the Korean current-version translation for the Go data types and templates guide.
- Reuse the existing Helm 3 Korean translation because the English source is unchanged.
- Polish the Korean prose conservatively while preserving technical terms and inline code.

## Validation

- `git diff --check`
- `yarn build --locale ko --no-minify`


TerryHowe (approved):
/lgtm

paigecalvert (changes requested):
@BusanGukbap thank you! could you also add the `default_lang_commit` to the front matter? It should be set to the latest commit that changed the corresponding English source file.

We are trying to add this to all localized pages to help us track drift with the English source. You can see details here: https://helm.sh/community/localization/#translating

TerryHowe (approved):
/lgtm

paigecalvert (approved):

PR #2206, opened 2026-08-01
fix(deps): bump uuid to 14.0.1 to patch CVE-2026-41907

TerryHowe:
## What

Adds a `resolutions` override pinning `uuid` to `14.0.1`.

## Why

The `uuid` package is pulled in transitively (`@docusaurus/core` → `webpack-dev-server` → `sockjs` → `uuid@^8.3.2`) and was resolving to `8.3.2`, which is affected by:

- **GHSA-w5hq-g745-h8pq** / **CVE-2026-41907** (medium): the `v3()`, `v5()`, and `v6()` API methods perform a **silent out-of-bounds write** when given a too-small `buf` or too-large `offset`, instead of throwing `RangeError` like `v4()`/`v1()`/`v7()` do.

Since `sockjs` requests `^8.3.2` (which can't reach the patched `11.1.1`+), a `resolutions` entry is used to force the patched version. `14.0.1` is the current latest and is outside all vulnerable ranges. Per the repo's guidance to pin security-critical deps to exact versions, it is pinned exactly.

## Testing

- `yarn install` — resolves `uuid` to `14.0.1` (`yarn why uuid` confirms sockjs now uses it)
- `yarn build --locale en` — builds successfully

Fixes Dependabot alert #93.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "serialize-javascript": ">=7.0.5",
     "path-to-regexp": "0.1.13",
     "minimatch": ">=3.1.3",
-    "browserslist": "^4.28.1"
+    "browserslist": "^4.28.1",
+    "uuid": "14.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


paigecalvert (approved):
/lgtm

PR #2211, opened 2026-08-09
docs: warn about YAML injection in chart templates

### LINKED ISSUE #2154: Document how to prevent YAML-injection in Chart Templates?

locker95:
The template guide shows `quote` / `toYaml` / `nindent` in places, but there wasn't a single warning about what happens if you skip them. Added a short section under YAML techniques (with a pointer from the intro) on safe embedding of values.

Fixes #2154

### DIFF
--- a/docs/chart_template_guide/yaml_techniques.md
+++ b/docs/chart_template_guide/yaml_techniques.md
@@ -374,7 +374,10 @@ the rendered YAML. This is called _YAML injection_.
 YAML injection can not only break installs, but can
 also inject extra keys into a Kubernetes manifest.
 
-Prefer helpers that encode or structure the data for you:
+### Best practices
+
+To avoid YAML injection when inserting Helm values,
+prefer helpers that encode or structure the data for you, as shown in the following table:
 
 | Goal | Prefer | Avoid |
 | ---- | ------ | ----- |
@@ -383,7 +386,16 @@ Prefer helpers that encode or structure the data for you:
 | Embed maps / lists | `{{ toYaml .Values.extraEnv \| nindent 8 }}`. `nindent` only adds indentation; it doesn't encode YAML. Pair it with `toYaml` for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an untrusted string is still injectable. | `{{ .Values.extraEnv \| nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
 | Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
 
-Example — safe nested object (map → YAML, then indent):
+The following describes general best practices for preventing YAML injection when inserting values:
+
+- Treat every `{{ ... }}` expression that is rendered inside YAML as untrusted input.
+- Use `quote`, `toYaml`, and `nindent`/`indent` together to preserve YAML structure. Avoid manually constructing YAML using string concatenation.
+- Run `helm template` and schema validation in CI so malformed values fail
+  before they reach the cluster.
+
+### Examples
+
+#### Safe nested object (map to YAML, then indent)
 
 ```yaml
 apiVersion: v1
@@ -396,7 +408,7 @@ data:
 {{- toYaml .Values.config | nindent 4 }}
 ```
 
-Example — unsafe patterns:
+#### Unsafe patterns
 
 ```yaml
 # BAD: a value of "x\n  evil: true" becomes an extra key
@@ -408,14 +420,6 @@ data:
   app.conf: |
 {{ .Values.appConf | nindent 4 }}
 ```
-
-When in doubt:
-
-1. Treat every `{{ ... }}` that lands inside YAML structure as untrusted input.
-2. Use `quote`, `toYaml`, and `nindent`/`indent` together rather than string
-   concatenation.
-3. Run `helm template` (and schema validation) in CI so malformed values fail
-   before they reach the cluster.
 
 For more information, see [Template Functions and Pipelines](functions_and_pipelines.mdx),
 [Indenting and Templates](#indenting-and-templates) on this page, and the
--- a/docs/chart_template_guide/yaml_techniques.md
+++ b/docs/chart_template_guide/yaml_techniques.md
@@ -382,9 +382,8 @@ prefer helpers that encode or structure the data for you, as shown in the follow
 | Goal | Prefer | Avoid |
 | ---- | ------ | ----- |
 | Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
-| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config \| nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
-| Embed maps / lists | `{{ toYaml .Values.extraEnv \| nindent 8 }}`. `nindent` only adds indentation; it doesn't encode YAML. Pair it with `toYaml` for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an untrusted string is still injectable. | `{{ .Values.extraEnv \| nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
-| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
+| Insert a string as one YAML scalar | `quote`, or a block scalar header (`\|` / `>`) plus `{{- .Values.config \| nindent N }}` where `N` is greater than the parent indent (usually parent + 2) | Bare `{{ .Values.config }}` or `nindent` without a block header — a newline in the value can start a new key |
+| Embed maps / lists (collections) | `{{ toYaml .Values.extraEnv \| nindent N }}` so `toYaml` emits the collection and `nindent` only shifts it | `{{ .Values.extraEnv \| nindent N }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
 
 The following describes general best practices for preventing YAML injection when inserting values:
 
@@ -401,11 +400,20 @@ The following describes general best practices for preventing YAML injection whe
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "mychart.fullname" . }}
+  name: {{ include "mychart.fullname" . | quote }}
 data:
   # values.config is a map; toYaml emits nested YAML, nindent only indents it
   config.yaml: |
-{{- toYaml .Values.config | nindent 4 }}
+    {{- toYaml .Values.config | nindent 4 }}
+```
+
+A string under a block scalar is the same idea — keep the `{{-` so you do
+not get an extra blank line, and pick `N` larger than the parent indent:
+
+```yaml
+data:
+  app.conf: |
+    {{- .Values.appConf | nindent 4 }}
 ```
 
 #### Unsafe patterns
@@ -415,10 +423,9 @@ data:
 data:
   app.conf: {{ .Values.appConf }}
 
-# BAD: nindent alone still injects — it only adds spaces
+# BAD: nindent without a block scalar header only adds spaces
 data:
-  app.conf: |
-{{ .Values.appConf | nindent 4 }}
+  app.conf: {{ .Values.appConf | nindent 4 }}
 ```
 
 For more information, see [Template Functions and Pipelines](functions_and_pipelines.mdx),
--- a/docs/chart_template_guide/yaml_techniques.md
+++ b/docs/chart_template_guide/yaml_techniques.md
@@ -9,6 +9,10 @@ we'll look at the YAML format. YAML has some useful features that we, as
 template authors, can use to make our templates less error prone and easier to
 read.
 
+When you embed chart values into manifests, also read
+[Safe embedding of values (YAML injection)](#safe-embedding-of-values-yaml-injection)
+so user-supplied strings cannot reshape the document.
+
 ## Scalars and Collections
 
 According to the [YAML spec](https://yaml.org/spec/1.2/spec.html), there are two
@@ -360,3 +364,62 @@ use [library charts](/topics/library_charts.md) instead of YAML anchors.
 Library charts are designed for reuse,
 and aren't subject to the round-trip pitfall described above.
 :::
+
+## Safe embedding of values (YAML injection)
+
+Chart values often come from untrusted or multi-tenant sources. If you splice a
+value into a template with raw `{{ .Values.foo }}`, a value that contains
+newlines, `:` mapping indicators, or list markers can change the structure of
+the rendered YAML (YAML injection). That can break installs—or worse, inject
+extra keys into a Kubernetes manifest.
+
+Prefer helpers that encode or structure the data for you:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name | quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config | nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv | nindent 8 }}` | `{{ .Values.extraEnv | nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
+
+`nindent` only adds indentation; it does **not** encode YAML. Pair it with `toYaml`
+for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an
+untrusted string is still injectable.
+
+Example — safe nested object (map → YAML, then indent):
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mychart.fullname" . }}
+data:
+  # values.config is a map; toYaml emits nested YAML, nindent only indents it
+  config.yaml: |
+{{- toYaml .Values.config | nindent 4 }}
+```
+
+Example — unsafe patterns:
+
+```yaml
+# BAD: a value of "x\n  evil: true" becomes an extra key
+data:
+  app.conf: {{ .Values.appConf }}
+
+# BAD: nindent alone still injects — it only adds spaces
+data:
+  app.conf: |
+{{ .Values.appConf | nindent 4 }}
+```
+
+When in doubt:
+
+1. Treat every `{{ ... }}` that lands inside YAML structure as untrusted input.
+2. Use `quote`, `toYaml`, and `nindent`/`indent` together rather than string
+   concatenation.
+3. Run `helm template` (and schema validation) in CI so malformed values fail
+   before they reach the cluster.
+
+See also [Functions and Pipelines](functions_and_pipelines.mdx),
+[Indenting and Templates](#indenting-and-templates), and the
+[function list](function_list.mdx) (`toYaml`, `quote`, `nindent`).
--- a/docs/chart_template_guide/yaml_techniques.md
+++ b/docs/chart_template_guide/yaml_techniques.md
@@ -9,8 +9,8 @@ we'll look at the YAML format. YAML has some useful features that we, as
 template authors, can use to make our templates less error prone and easier to
 read.
 
-When you embed chart values into manifests, also read
-[Safe embedding of values (YAML injection)](#safe-embedding-of-values-yaml-injection)
+When you insert chart values into manifests, also read
+[Prevent YAML injection when inserting Helm values](#prevent-yaml-injection-when-inserting-helm-values)
 so user-supplied strings cannot reshape the document.
 
 ## Scalars and Collections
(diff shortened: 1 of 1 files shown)


docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,53 @@ coffees:
...
+newlines, `:` mapping indicators, or list markers can change the structure of
+the rendered YAML (YAML injection). That can break installs—or worse, inject
+extra keys into a Kubernetes manifest.
+
+Prefer helpers that encode or structure the data for you:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
+| Indent a multi-line string | `{{ .Values.config \| nindent 4 }}` or `\| indent 4` after `toYaml` | Pasting multi-line values next to a key without indentation |
```
TerryHowe: This `{{ .Values.config \| nindent 4 }}` is still injectable.

TerryHowe (changes requested):

locker95:
Good catch — `nindent` only indents. I updated the table so maps/lists go through `toYaml` first and called out that bare `nindent` on free-form values is still injectable.

locker95:
you're right — bare `nindent` doesn't encode anything. dropped it as a safe preference and called that out with a bad example.

locker95:
@TerryHowe — rebased the table so bare nindent isn't listed as safe anymore (maps go through toYaml first, and there's a bad example calling out that nindent alone still injects). PTAL when you get a chance.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+extra keys into a Kubernetes manifest.
+
+Prefer helpers that encode or structure the data for you:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name | quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config | nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv | nindent 8 }}` | `{{ .Values.extraEnv | nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
```
paigecalvert: It doesn't look like this table is rendering properly. Need to escape the | chars?

Image

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
 
 Because Helm and Kubernetes often read, modify, and then rewrite YAML files, the
 anchors will be lost.
+
+## Safe embedding of values (YAML injection)
```
paigecalvert:
```suggestion
## Prevent YAML injection when inserting Helm values
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+
+Chart values often come from untrusted or multi-tenant sources. If you splice a
+value into a template with raw `{{ .Values.foo }}`, a value that contains
+newlines, `:` mapping indicators, or list markers can change the structure of
+the rendered YAML (YAML injection). That can break installs—or worse, inject
+extra keys into a Kubernetes manifest.
```
paigecalvert:
```suggestion
the rendered YAML. This is called _YAML injection_.
YAML injection can not only break installs, but can
also inject extra keys into a Kubernetes manifest.
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name | quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config | nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv | nindent 8 }}` | `{{ .Values.extraEnv | nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
+
+`nindent` only adds indentation; it does **not** encode YAML. Pair it with `toYaml`
+for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an
+untrusted string is still injectable.
```
paigecalvert: This is a good explainer, but it should probably be moved next to "`toYaml` + `nindent` (or chart helpers)" in the table, since I think it's specific to that row.

Alternatively, we could also add a Why? column so that there is a clear explanation of each recommendation.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+Prefer helpers that encode or structure the data for you:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name | quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config | nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv | nindent 8 }}` | `{{ .Values.extraEnv | nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
+
+`nindent` only adds indentation; it does **not** encode YAML. Pair it with `toYaml`
```
paigecalvert:
```suggestion
`nindent` only adds indentation; it doesn't encode YAML. Pair it with `toYaml`
```
(style edit: don't use bold text for emphasis)

docs/chart_template_guide/yaml_techniques.md
paigecalvert: WDYT about this heading edit? Wanted to address a few things:
- change from passive to active voice
- "(YAML injection)" in parentheses felt a little vague. wanted to make it clear that it's something that you're trying to avoid
- "insert" felt like a plainer, more straightforward verb for this vs "embed". Open to leaving it as embed if that feels more accurate, though

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+
+When in doubt:
+
+1. Treat every `{{ ... }}` that lands inside YAML structure as untrusted input.
+2. Use `quote`, `toYaml`, and `nindent`/`indent` together rather than string
+   concatenation.
+3. Run `helm template` (and schema validation) in CI so malformed values fail
+   before they reach the cluster.
+
+See also [Functions and Pipelines](functions_and_pipelines.mdx),
```
paigecalvert:
```suggestion
For more information, see [Template Functions and Pipelines](functions_and_pipelines.mdx),
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+[Indenting and Templates](#indenting-and-templates), and the
```
paigecalvert:
```suggestion
[Indenting and Templates](#indenting-and-templates) on this page, and the
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,62 @@ coffees:
...
+[function list](function_list.mdx) (`toYaml`, `quote`, `nindent`).
```
paigecalvert:
```suggestion
[Template Function List](function_list.mdx) (`toYaml`, `quote`, `nindent`).
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -9,6 +9,10 @@ we'll look at the YAML format. YAML has some useful features that we, as
 template authors, can use to make our templates less error prone and easier to
 read.
 
+When you embed chart values into manifests, also read
+[Safe embedding of values (YAML injection)](#safe-embedding-of-values-yaml-injection)
```
paigecalvert: note: if we change the section heading below this needs to be updated accordingly

paigecalvert (changes requested):
@locker95  a few comments. main issue was table rendering seems off!

locker95:
escaped the table pipes so it actually renders, moved the nindent note into that row, and took your heading/wording/link edits.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,59 @@ coffees:
...
+## Prevent YAML injection when inserting Helm values
+
+Chart values often come from untrusted or multi-tenant sources. If you splice a
+value into a template with raw `{{ .Values.foo }}`, a value that contains
+newlines, `:` mapping indicators, or list markers can change the structure of
+the rendered YAML. This is called _YAML injection_.
+YAML injection can not only break installs, but can
+also inject extra keys into a Kubernetes manifest.
+
+Prefer helpers that encode or structure the data for you:
```
paigecalvert:
```suggestion
### Best practices

To avoid YAML injection when inserting Helm values,
prefer helpers that encode or structure the data for you, as shown in the following table:
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,59 @@ coffees:
...
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config \| nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv \| nindent 8 }}`. `nindent` only adds indentation; it doesn't encode YAML. Pair it with `toYaml` for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an untrusted string is still injectable. | `{{ .Values.extraEnv \| nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
+
```
paigecalvert, line 387:
```suggestion
The following describes general best practices for preventing YAML injection when inserting values:

- Treat every `{{ ... }}` expression that is rendered inside YAML as untrusted input.
- Use `quote`, `toYaml`, and `nindent`/`indent` together to preserve YAML structure. Avoid manually constructing YAML using string concatenation.
- Run `helm template` and schema validation in CI so malformed values fail
   before they reach the cluster.
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,59 @@ coffees:
...
+Example — safe nested object (map → YAML, then indent):
```
paigecalvert:
```suggestion
### Examples

#### Safe nested object (map to YAML, then indent)
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,59 @@ coffees:
...
+kind: ConfigMap
+metadata:
+  name: {{ include "mychart.fullname" . }}
+data:
+  # values.config is a map; toYaml emits nested YAML, nindent only indents it
+  config.yaml: |
+{{- toYaml .Values.config | nindent 4 }}
+```
+
+Example — unsafe patterns:
```
paigecalvert:
```suggestion
#### Unsafe patterns
```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,59 @@ coffees:
...
+```
+
+When in doubt:
+
+1. Treat every `{{ ... }}` that lands inside YAML structure as untrusted input.
+2. Use `quote`, `toYaml`, and `nindent`/`indent` together rather than string
+   concatenation.
+3. Run `helm template` (and schema validation) in CI so malformed values fail
+   before they reach the cluster.
+
```
paigecalvert:
```suggestion
```

paigecalvert: Recommend moving this up next to the table, which is also describing best practices

paigecalvert (commented):
thank you @locker95 ! I did one more small pass of edits; just copyedits and also shuffled the info around a bit so that all the "best practices" guidance was grouped together.

locker95:
took the headings/shuffle — best practices sit with the table now, examples underneath.

paigecalvert (approved):
/lgtm

locker95:
@TerryHowe the nindent/table bits from your review should be in — paige did a later edit pass as well. ptal when you have a minute?

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,63 @@ coffees:
...
+### Best practices
+
+To avoid YAML injection when inserting Helm values,
+prefer helpers that encode or structure the data for you, as shown in the following table:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config \| nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
```
meeque: Not sure what you mean by "multi-line free-form string"? This term is not used in the latest YAML spec.

YAML does have different mechanisms to express string scalars. These are:

* [flow style:](https://yaml.org/spec/1.2.2/#73-flow-scalar-styles)
  * double-quoted
  * single-quoted
  * plain
* [block style](https://yaml.org/spec/1.2.2/#81-block-scalar-styles)
  * literal
  * folded

I'm assuming that you are referring to one of the block styles, because these are the only ones that are typically written across multiple lines. (Yes, the flow styles can also be written across multiple lines, but the rules fore this are rather complex. I haven't seen this approach much, and I wouldn't recommend it.)

In either case, these are all string **scalars**. In contrast, you're mentioning **map/list** types in the "Prefer" column. These are **collections** (i.e. mappings or sequences) in YAML terminology, which are distinct from strings. I think the current phrasing of this whole row may confuse readers.

Btw, the YAML parser distinguishes between block scalars and block collections by [block scalar headers](https://yaml.org/spec/1.2.2/#811-block-scalar-headers). This involves either a `|` or a `>` character. If one of these headers is present, the YAML parser always treats the following block as a **scalar**, otherwise it treats it as a **collection**.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,63 @@ coffees:
...
+
+```yaml
+# BAD: a value of "x\n  evil: true" becomes an extra key
+data:
+  app.conf: {{ .Values.appConf }}
+
+# BAD: nindent alone still injects — it only adds spaces
+data:
+  app.conf: |
+{{ .Values.appConf | nindent 4 }}
```
meeque: Not sure about this. Have you tested it?

Since `nindent` adds spaces to *each line* of the value, this is actually sufficient to prevent injection.
However there is some nuance to this:

* The previous line contains a block scalar header, namely `|` in this example.
  If this were absent, this whole construct would be vulnerable to YAML injection.
* The `nindent` is called with the correct number, namely `4` in this example.
  If this were `2` or smaller, this whole construct might be vulnerable to YAML injection.

However, I'm not 100% certain about this, due to the **extra line-break** in this example. We have 2 leading line-breaks here, one from the template itself, one from `nindent`. This may or may not work reliably.
In either case, I would recommend either of these alternatives instead:

* Use go template's `{{-` which removes all preceding white-space, including the line break.
  The leading white-space before `{{-` is is optional, but may improve readability of the template code:
  ```
    app.conf: |
      {{- .Values.appConf | nindent 4 }}
  ```
* Use `indent` rather than `nindent`, which does not add an extra line-break.
  Here, it is essential that there is no additional indent before the `{{`.
  Imho this makes the template code less readable:
  ```
    app.conf: |
  {{ .Values.appConf | indent 4 }}
  ```
* Turn it into a one-liner.
  Remove the line-break from the template, but retain the line-break that `nindent` introduces.
  Could be confusing to some users, because a YAML block scalar header is not followed by content on the same line:
  ```
    app.conf: | {{ .Values.appConf | nindent 4 }}
  ```

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,63 @@ coffees:
...
+### Best practices
+
+To avoid YAML injection when inserting Helm values,
+prefer helpers that encode or structure the data for you, as shown in the following table:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
+| Multi-line free-form string | Store structured data as a map/list and use `toYaml`, or treat the whole value as one quoted scalar with `quote` | `{{ .Values.config \| nindent 4 }}` — `nindent` only indents; newline-bearing values can still inject keys |
+| Embed maps / lists | `{{ toYaml .Values.extraEnv \| nindent 8 }}`. `nindent` only adds indentation; it doesn't encode YAML. Pair it with `toYaml` for maps/lists, or use `quote` for a single scalar. Bare `nindent` on an untrusted string is still injectable. | `{{ .Values.extraEnv \| nindent 8 }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
```
meeque: I'm missing some context here. Why is it `nindent 8` rather than `nindent 6` or `nindent 23`?

Of course, this depends on the indent of the surrounding YAML code in the template. If too little indent is used, this could still result in YAML injection. It should be documented how much indent is secure!
Imho this is at least `1` more than the indent of the surrounding YAML block. Though `2` more than the indent of the surrounding YAML block is commonly used, which improves readability of the generated YAML code.

Imho, this line also mixes statements about **YAML collections** and **YAML scalars** in a confusing manner. See my comment on the previous table row.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,63 @@ coffees:
...
+| Labels / annotations maps | `toYaml` + `nindent` (or chart helpers) | Concatenating free-form label lines from values |
```
meeque: Not sure, if these deserve special mention?

From a YAML perspective, these are equivalent to other uses of **collection types**. Maybe add labels and annotations to the below code examples instead?

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -353,3 +357,63 @@ coffees:
...
+
+### Examples
+
+#### Safe nested object (map to YAML, then indent)
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mychart.fullname" . }}
```
meeque: Does the the `include` reference a standard template here?

If not, the template code here should not make any assumptions about what that included template generates. Instead, it should follow the above "best practices"!

In this example, the following might be most appropriate:
```
  name: {{ include "mychart.fullname" . | quote }}
```

Btw, the existing [docs for for the `include` function ](https://helm.sh/docs/chart_template_guide/named_templates#the-include-function) already get this right. Though the context is different in the examples there, so these use `nindent` instead of `quote`.

docs/chart_template_guide/yaml_techniques.md
meeque: I'd say this depends on the context. Imho this is perfectly secure, if this is used inside a YAML block whose indent is smaller than 4. See my new comment there...

meeque:
Hi all!

Sorry, for meddling again. I really appreciate that you are taking YAML injection seriously. Frankly, I've discussed this topic elsewhere in the past, and some people didn't even acknowledge that there is a problem here.

However, are you aware that there are two other PRs for the leading issue already? See #2158 and #2165. The latter one is from me. And I understand that the changes that it proposes may be too far reaching. I guess that I'm not involved in Helm enough to propose such bold changes.

But could you at least **cross-check the contents** of this PR with the other two? Frankly, I fear that some of the new docs in this PR are incorrect. Or, at least, misleading and incomplete. See my comments above.

May I also suggest to have **test cases** for all the new advice and examples? I guess that tests may not fit into this *helm-www* project nicely. But YAML syntax is really tricky and imho template code examples should be covered by tests!

I have this [helm-charts-yaml-injection](meeque/helm-charts-yaml-injection) project (I've mentioned it earlier) that tests all the advice that I've given on this topic in the past. If you think that it's missing a relevant test case, feel free to reach out to me. I'll happily add anything that might be relevant!

locker95:
switched the table to scalar vs collection, quoted the include, and showed `{{- nindent N }}` under a block header with N bigger than the parent indent.

locker95:
rebased onto main — the conflict was just the new library-charts note above the injection section.

docs/chart_template_guide/yaml_techniques.md
```diff
@@ -360,3 +364,70 @@ use [library charts](/topics/library_charts.md) instead of YAML anchors.
...
+### Best practices
+
+To avoid YAML injection when inserting Helm values,
+prefer helpers that encode or structure the data for you, as shown in the following table:
+
+| Goal | Prefer | Avoid |
+| ---- | ------ | ----- |
+| Quote a string scalar | `{{ .Values.name \| quote }}` | `{{ .Values.name }}` in a bare field |
+| Insert a string as one YAML scalar | `quote`, or a block scalar header (`\|` / `>`) plus `{{- .Values.config \| nindent N }}` where `N` is greater than the parent indent (usually parent + 2) | Bare `{{ .Values.config }}` or `nindent` without a block header — a newline in the value can start a new key |
+| Embed maps / lists (collections) | `{{ toYaml .Values.extraEnv \| nindent N }}` so `toYaml` emits the collection and `nindent` only shifts it | `{{ .Values.extraEnv \| nindent N }}` without `toYaml`, or hand-rolled `key: {{ . }}` loops |
```
TerryHowe:
```suggestion
| Embed maps / lists (collections) | `{{ toYaml .Values.extraEnv \| nindent N }}` where `N` is greater than the parent indent (usually parent + 2), so `toYaml` emits the collection and `nindent` only shifts it into place | `{{ .Values.extraEnv \| nindent N }}` without `toYaml`, an `N` that is not greater than the parent indent, or hand-rolled `key: {{ . }}` loops |
```

TerryHowe (approved):
/lgtm

Issue #2220, opened 2026-08-18
Set onBrokenAnchors: "throw" so broken anchors fail CI


TerryHowe:
`docusaurus.config.js:66` sets:

```js
onBrokenLinks: "throw",
```

but does not set `onBrokenAnchors`, which defaults to `"warn"` in Docusaurus 3.x (this site is on 3.10.1). Broken *page* links fail the build; broken *anchors* only emit a warning that is easy to miss in Netlify output.

That is why the four broken French anchors in #2217 passed CI and the deploy preview.

Setting `onBrokenAnchors: "throw"` would catch this class of regression at PR time — valuable here specifically, because the ongoing work to add English anchor IDs to translated headings (see [`community/localization.md`](https://github.com/helm/helm-www/blob/main/community/localization.md)) changes heading IDs across many locales, and every such change can silently orphan an in-page link.

### Scope caveat

This is not a one-line change. A rough sweep of `main` finds roughly **95 already-broken same-page anchors** across the repo, which would all need fixing before the setting can be flipped:

| locale | approx. broken anchors |
|---|---|
| `ja` | 48 |
| `uk` | 32 |
| `ko` | 6 |
| `fr` | 3 (tracked in #2217) |
| English sources | 2 |
| `el`, `ru`, `pt`, `zh` | 1 each |

These counts are approximate — they come from a quick script that does not replicate every edge case of Docusaurus's slugger (duplicate-heading `-1` suffixes, some non-ASCII handling), so the real figure will differ somewhat. Running `yarn build` and collecting the actual anchor warnings is the reliable way to get the true list.

Suggested sequencing:

1. Capture the real warning list from `yarn build`.
2. Fix per locale, likely one issue or PR each.
3. Flip `onBrokenAnchors` to `"throw"`.

Filing this as a tracking issue rather than a ready-to-take task. If the localization working group is already tracking anchor cleanup somewhere, this should probably fold into that instead.

FaithKovi:
I got the warning list from `yarn build` and extracted the `Broken Anchor` warning with the following commands:
```
yarn build 2>&1 | tee build.log
grep -A2 "Broken anchor" build.log 
```

Here's the output:

```
- Broken anchor on source page path = /docs/intro/install:
   -> linking to /docs/intro/quickstart#initialize-a-helm-chart-repository

--
- Broken anchor on source page path = /de/blog/path-to-helm-v4:
   -> linking to /de/community/release_policy#major-releases
[...]

PR #2230, opened 2026-08-31
fix(i18n): fix zh locale broken anchors

FaithKovi:
This PR fixes the remaining broken anchors in the `zh` locale. 

All broken links already target English anchor IDs; the translated headings were missing the explicit IDs. Added `{#english-source-slug}` to each.

**Verification:** 
`yarn build --locale zh` had these 5 broken anchor warnings before (one was the shared `install → quickstart` anchor already fixed on `main`).

Refs #2220.

---

**Unrelated observation while verifying this PR:** several `zh` v3 pages use hardcoded absolute links with the locale prefix (e.g. `/zh/docs/topics/charts/`) instead of the file-relative form (`./charts.md`) used elsewhere in the repo.

These resolve fine in the full multi-locale build (CI/Netlify) and when I temporary set `onBrokenLinks: "warn"` locally, so nothing is broken. But without that, they fail in single-locale builds (`yarn build --locale zh`) locally, and because they bypass Docusaurus link resolution, they'd 404 silently rather than fail the build if a target page ever moves. I found 12 such links across 5 files while running the verification build above. See image below:

Screenshot 2026-08-31 at 08 42 26

Happy to open a separate issue/PR to convert them to relative links if that's welcome. I just didn't want to mix it into this one.

### DIFF
--- a/i18n/zh/docusaurus-plugin-content-docs/version-3/chart_template_guide/functions_and_pipelines.md
+++ b/i18n/zh/docusaurus-plugin-content-docs/version-3/chart_template_guide/functions_and_pipelines.md
@@ -158,7 +158,7 @@ drink: {{ .Values.favorite.drink | default (printf "%s-tea" (include "fullname" 
 模板函数和管道符是转换信息然后将其插入到YAML中的强有力方式。但是有些时候我们需要插入一些内容之前进行一些逻辑判断，而不仅仅是插入一个字符串。
 下一章节，我们会看到模板语言提供的控制结构。
 
-## 使用`lookup`函数
+## 使用`lookup`函数 {#using-the-lookup-function}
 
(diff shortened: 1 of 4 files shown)


TerryHowe (approved):
/lgtm

FaithKovi:
@TerryHowe please let me know what you think about the image attached.

TerryHowe:
Thanks for flagging the absolute-link issue separately rather than folding it in here — filed as #2236 with the full list.

Confirmed your finding: 13 links across the 5 files you identified, and `zh` is the only locale with the pattern. I also turned up 2 related links in `zh` that are absolute *without* the locale prefix (the mirror-image problem), so those are captured in the issue too. A PR converting them to file-relative links would be very welcome.

Issue #2217, opened 2026-08-18
Broken anchor links in French docs after the fr anchor-ID migration


TerryHowe:
PR #2216 added English anchor IDs to French headings (per the convention in [`community/localization.md`](https://github.com/helm/helm-www/blob/main/community/localization.md)), but four in-page links still point at the old auto-generated French slugs, or were retargeted to an ID that was never added. These render as dead links.

| File | Link | Actual heading |
|---|---|---|
| `i18n/fr/.../chart_template_guide/accessing_files.md:145` | `[Glob](#motifs-glob)` | `:102` — `## Motifs glob {#glob-patterns}` |
| `i18n/fr/.../topics/plugins.md:323` | `[Complétion dynamique](#complétion-dynamique)` | `:372` — `### Complétion dynamique {#dynamic-completion}` |
| `i18n/fr/.../topics/plugins.md:409` | `[Complétion dynamique](#complétion-dynamique)` | `:372` — same heading |
| `i18n/fr/.../topics/charts.md:453` | `[child-parent](#using-the-child-parent-format)` | `:496` — `##### Utiliser le format child-parent` (no explicit ID) |

The first three are fixed by updating the link to the new English anchor ID.

The fourth is the reverse case: the link was changed to `#using-the-child-parent-format`, but the heading never received that anchor. Either add `{#using-the-child-parent-format}` to line 496 (preferred — it matches the English source slug and follows the convention) or revert the link to `#utiliser-le-format-child-parent`.

A third occurrence in `plugins.md` *was* correctly updated to `#dynamic-completion`, so these look like misses rather than intent.

### Verified against

`main` at e231423af18f4795dcf5e94107a1e9ed0e0deb83.

Note that these do not fail the build today — see the companion issue on `onBrokenAnchors`.

PR #2231, opened 2026-08-31
fix(i18n): fix fr locale broken anchors

FaithKovi:
This PR fixes the broken anchors in the `fr` locale (follow-up to #2217/#2225, which covered the same-page cases). 

All four links already target English anchor IDs; the translated headings were missing the explicit IDs. Added `{#english-source-slug}` to each of them.

**Verification:** 
`yarn build --locale fr`,  0 broken anchors after.

Refs #2220.


TerryHowe (approved):
/lgtm

PR #2232, opened 2026-08-31
fix(i18n): fix es locale broken anchors

FaithKovi:
This PR fixes the broken anchors in the `es` locale. 

All three links already target English anchor IDs; the translated headings were missing the explicit IDs. Added `{#english-source-slug}`.

**Verification:** 
`yarn build --locale es` had 0 broken anchors after

Refs #2220.


TerryHowe (approved):
/lgtm

PR #2235, opened 2026-08-31
fix(i18n): fix ja locale broken anchors

FaithKovi:
This PR fixes the broken anchors in the `ja` locale. 

All broken links already target English anchor IDs; the translated headings were missing the explicit IDs. Added `{#english-source-slug}`.

Only headings with a broken inbound anchor received IDs; headings whose auto-generated slugs already match the English ones are untouched.

**Verification:** 
`yarn build --locale ja` has 0 broken anchors after. 

Refs #2220.


TerryHowe (approved):
/lgtm

PR #2238, opened 2026-08-31
fix(i18n): resolve zh docs links relatively

### LINKED ISSUE #2236: Convert hardcoded absolute links in zh docs to file-relative links
The `zh` v3 docs use hardcoded absolute links carrying the locale prefix (e.g. `[Charts](/zh/docs/topics/charts/)`) instead of the file-relative form (`[Charts](./charts.md)`) used everywhere else in the repo.

Reported by @FaithKovi while verifying #2230; filing separately so it isn't mixed into an anchor-fix PR.

### Why it matters

These render correctly today in the full multi-locale build (CI and Netlify), so nothing is broken for readers right now. Two problems all the same:

1. **They break single-locale builds.** `yarn build --locale zh` emits the `zh` site at the root `baseUrl`, so the `/zh/` prefix no longer exists and every one of these links is a broken link. With `onBrokenLinks: "throw"` that fails the build, which makes single-locale builds unusable as a verification step for `zh` — exactly the workflow contributors need while fixing anchors for #2220.
2. **They bypass Docusaurus link resolution.** File-relative `.md` links are resolved and validated at build time, so a moved or renamed target fails the build. A hardcoded URL string is never checked — if a target page moves, these silently 404 in production instead.

### Scope

13 links across 5 files, all under `i18n/zh/docusaurus-plugin-content-docs/version-3/`:

| file | count |
|---|---|
| `topics/chart_repository.md` | 6 |
| `howto/chart_repository_sync_example.md` | 2 |
| `topics/advanced.md` | 2 |
| `topics/chart_tests.md` | 2 |
| `topics/charts_hooks.md` | 1 |

Reproduce with:

```sh
git grep -n -E '\]\(/zh/' -- 'i18n/zh/**'
```

`zh` is the only locale affected — the same grep against every other locale prefix returns nothing.

### Related variant

Two further links in `zh` are absolute but *without* the locale prefix, which has the mirror-image problem (they resolve in a single-locale build but point at the English page in the multi-locale build):

- `i18n/zh/docusaurus-plugin-content-docs-community/current/developers.md:60` → `/docs/helm`
- `i18n/zh/docusaurus-plugin-content-docs/version-3/chart_template_guide/getting_started.md:216` → `/docs/chart_template_guide/index.mdx`

The second is also linking to a source path rather than a route. Worth folding into the same cleanup.

### Suggested fix

Convert all 15 to file-relative `.md` / `.mdx` links pointing at the sibling `zh` source files, then confirm with `yarn build --locale zh`.

Refs #2220.

Soumoditya:
## Summary

Fixes the unresolved Helm CLI reference in the Chinese developer guide. The link now targets the published Helm 3 documentation route.

Closes #2236

## Validation

- `git diff --check`
- `yarn build --locale zh` (passed)

### DIFF
--- a/i18n/zh/docusaurus-plugin-content-docs-community/current/developers.md
+++ b/i18n/zh/docusaurus-plugin-content-docs-community/current/developers.md
@@ -57,7 +57,7 @@ Go依赖管理在不断变化，而且在Helm生命周期中很可能发生变�
 从Helm 3开始，文档已经移动到了它自己的仓库中。当编制新特性时，请编写随附文档并提交到
 [helm-www](https://github.com/helm/helm-www) 仓库。
 
-有个例外：[Helm CLI 输出 (英文)](../../docusaurus-plugin-content-docs/version-3/helm/helm.md) 是 `helm` 程序自己生成的。
+有个例外：[Helm CLI 输出 (英文)](/docs/v3/helm/helm) 是 `helm` 程序自己生成的。
 查看 [更新Helm CLI参考文档](https://github.com/helm/helm-www#updating-the-helm-cli-reference-docs)
 来了解如何生成该输出。翻译后，不会生成CLI输出，但可以在 `/content/<lang>/docs/helm` 中找到。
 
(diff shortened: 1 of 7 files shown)


TerryHowe:
Sorry, dup of earlier https://github.com/helm/helm-www/pull/2237 thanks anyway

PR #2239, opened 2026-09-01
fix(config): set onbrokenanchors to throw

### LINKED ISSUE #2220: Set onBrokenAnchors: "throw" so broken anchors fail CI

FaithKovi:
Sets `onBrokenAnchors: "throw"` so broken anchors fail the build the same way broken links already do. This PR fixes #2220.

### Why it's safe to flip now

All broken anchors surfaced by `yarn build` have been fixed in the preceding PRs (verified per-locale, all builds now at zero broken-anchor warnings):

### Verification

Full `yarn build` on this branch (all 12 locales): completes with **zero** broken-anchor warnings and exits successfully with the new setting. Going forward, any new broken anchor on same-page or cross-page in any locale — fails CI instead of scrolling by as a warning.

See image below for the log tail of the current `yarn build` output.
Screenshot 2026-09-01 at 11 39 29

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -64,6 +64,7 @@ const config = {
   projectName: "helm-www", // Usually your repo name.
 
   onBrokenLinks: "throw",
+  onBrokenAnchors: "throw",
 
   i18n: {
     defaultLocale: "en",


TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

